PR #296, opened 2015-01-23
Make CLANG IBs to use unmodified CMSSW stable branch

davidlt:
Currently pre-3.6 Clang is only available in DEVEL IBs. Until
merged to stable CMSDIST branch we should use CMSDIST next branch
with new Clang.

Also make sure that Clang IBs are using plain CMSSW without any
additional patches, i.e., CMSSW stable branch.

Signed-off-by: David Abdurachmanov <davidlt@cern.ch>


ktf:
We should have a few builds with this and then kill the _CLANG_X branch, BTW...

davidlt:
If you are happy with loosing the history of CLANG_X branch, I am OK.

ktf:
So long and thanks for all the fish.. ;)

ktf:
So long and thanks for all the fish..

PR #320, opened 2015-02-12
Always create a tag in cmsdist per IB

nclopezo:
Right now, if there are errors in the build and there release queue has not set ALWAYS_TAG_CMSSW then no tag in cmsdist will be created. 

This is why when there are failures in the build the cmsdist tags can't be seen. (https://cmssdt.cern.ch/SDT/html/showIB.html#CMSSW_7_2_X_2015-02-11-1400) This should always create a "default" tag per IB.


nclopezo:
@smuzaffar, @ktf

PR #329, opened 2015-02-18
PR Tests: test clang compilation and improve results reporting script

nclopezo:


run-pr-tests
```diff
@@ -396,26 +390,60 @@ for WF in ${WORKFLOWS_FOR_VALGRIND_TEST//,/ }; do
...
+NEED_CLANG_TEST=false
+if curl $CONFIG_MAP_URL | grep $RELEASE_QUEUE | grep PRS_TEST_CLANG= | grep SCRAM_ARCH=$ARCHITECTURE; then
+  NEED_CLANG_TEST=true
+fi
+
+if [ "X$TEST_CLANG_COMPILATION" = Xtrue -a $NEED_CLANG_TEST = true ]; then
+  $WORKSPACE/cms-bot/report-pull-request-results TESTS_RUNNING --pr $PULL_REQUEST_NUMBER -c $LAST_COMMIT --pr-job-id $PULL_REQUEST_JOB_ID --add-message "Testing Clang compilation" $DRY_RUN
+
+  #first, add the command to the log
+  echo "scram b vclean && scram build -k -j $(getconf _NPROCESSORS_ONLN) USER_CXXFLAGS='-fsyntax-only' COMPILER='llvm compile'" > $WORKSPACE/buildClang.log
```
ktf: Did you move to a different area compared to the gcc one? This will kill the gcc build otherwise.

nclopezo: I had done it so it is executed at the end of all the other tests. But I just changed it to be at the beginning. and then  it does scram b vclean

run-pr-tests
```diff
@@ -151,11 +153,48 @@ fi
...
+
+  scram b vclean && scram build -k -j $(getconf _NPROCESSORS_ONLN) USER_CXXFLAGS='-fsyntax-only' COMPILER='llvm compile' 2>&1 | tee -a $WORKSPACE/buildClang.log
```
nclopezo: ah I see that this is the most recent diff. Here it does scram b vclean. would it be ok like this? or should I put it back at the end?

run-pr-tests
```diff
@@ -151,11 +153,48 @@ fi
...
+else
+  echo 'CLANG_COMPILATION_RESULTS;NOTRUN' >> $RESULTS_FILE
 fi
 
-scram b -k -j $(getconf _NPROCESSORS_ONLN) 2>&1 | tee -a $WORKSPACE/build.log
+
+# #############################################
+# test compilation with GCC
+# ############################################
+scram b vclean && cram b -k -j $(getconf _NPROCESSORS_ONLN) 2>&1 | tee -a $WORKSPACE/build.log
```
nclopezo: Here it does scram b vclean again

run-pr-tests
ktf: Ok.

nclopezo:
I found an issue with the reporting of the results in Github. I will test it better and merge it tomorrow.

PR #335, opened 2015-02-20
Move CMSSW_7_5_X IBs to use IB/CMSSW_7_5_X/gcc481

ktf:
Was erroneously using `IB/CMSSW_7_5_X/stable` before. Hopefully this is the root cause of the CLHEP issues there.


ktf:
@davidlange6 @nclopezo @smuzaffar

PR #346, opened 2015-02-21
Allow "Upload all" to upload.

ktf:


ktf:
@davidlange6 this should get your release uploaded

PR #362, opened 2015-03-04
AutoBuilds: Allow to build cmssw-tool-conf before starting cmssw

nclopezo:


build-release
```diff
@@ -5,12 +5,20 @@
...
 QUEUE=$CMSSW_QUEUE
 
+if [ "X$ONLY_BUILD_TOOLCONF" = Xtrue ]; then
+  POST_ACTION='TOOL_CONF_BUILDING'
+else
+  POST_ACTION='BUILDING'
+fi
+
+
 git clone git@github.com:cms-sw/cmsdist.git $BUILD_DIR/CMSDIST
```
smuzaffar: what will happen if CMSDIST is already there? e.g. cloned by tool-conf process?

build-release
```diff
@@ -53,6 +61,19 @@ pushd $BUILD_DIR
...
+    pushd $HERE/$CMSSW_X_Y_Z-build/CMSDIST
+      git tag  REL/$CMSSW_X_Y_Z-tool-conf/$ARCHITECTURE
+      git push origin REL/$CMSSW_X_Y_Z-tool-conf/$ARCHITECTURE
+    popd
+    # If it reaches here it is because there were no errors
+    echo 'ALL_OK'
+    exit 0
+  fi
+
   $CMSBUILD_CMD deprecate-local $TOOL_CONF_PACKAGES
```
smuzaffar: Move this line above the check  i.e. just after the --sync-back upload.

build-release
nclopezo: It deletes the workspace before the build starts. And also it can choose a different machine for the build. Jenkins can assign cmsbuild13 to build tool-conf and then cmsbuil15 to build cmssw.

build-release
nclopezo: Ok I did here: https://github.com/cms-sw/cms-bot/commit/67f2f6a0ac21bfbd3e9e2d29ec63912a446eec0b#diff-e5ce8e5fc379eb81997d1448232a651e

build-release
smuzaffar: ah OK ..... I thought we are going to reuse the build area created by 
the tool-conf (to avoid downloading the RPMs again). It is true that if 
we reuse the build area then a release build might have to wait as the 
machine might be busy building ib or validation.
So who is going to cleanup the tool-conf build area?

nclopezo: There was a similar issue for the upload. The upload is run in mesos, it ssh the machine that did the build and starts the upload. 
And I will add the tool-conf area to the cleanup.  The cleanup looks in the issue for which were the machines used for the build, now it can also look for the machines that did the tool-conf build and trigger a cleanup for them.

PR #410, opened 2015-03-25
update validate.C: works with ROOT6, added new variables

slava77:
validate.C and compareValHists.C  needed an update to compile and run with ROOT6

I've updated from their original location: https://github.com/slava77/cms-reco-tools


slava77:
without the fixes the validateJR in 75X is  broken

slava77:
the fix is backward compatible

slava77:
uhm, validate.C now compiles but still doesn't run properly (specific to ROOT6), investigating ...

PR #420, opened 2015-03-31
IB builds: Check also if repository is empty

nclopezo:
Before, if the repository directory existed but it was empty it didn't clone it again. Now it should also check if it is empty.


nclopezo:
@ktf

PR #441, opened 2015-04-28
IBPages: Allow to set the artifacts machine on the parameters

nclopezo:


ktf:
Maybe you want to use an option, rather than a positional argument? Something like "-t <machine>". It would be interesting if we could use EOS as well for this.

PR #446, opened 2015-05-06
A set of scripts to deploy WMAgent and run the WMCore unit test

ericvaandering:
This set of scripts gets me through my first Jenkins job. I'd appreciate a review, of course, since this is new to me.


DMWM/activate-start-agent.sh
```diff
@@ -0,0 +1,8 @@
+#! /bin/sh
```
nclopezo: You can add "-ex" to immediately terminate after a command finishes with an error, and to show the command that is being executed.

PR #485, opened 2015-06-04
EventFilter/Utilities added to reco as well

slava77:
@davidlange6 please confirm


slava77:
@degano @smuzaffar please check and possibly merge

PR #487, opened 2015-06-05
update comparisons workflow map for short workflow names

slava77:
- map file updates for fwlite-based comparisons:
    - (some updated from Carl) added 1307.0; fixed file name in 1003.0;; 
    - added a new option name for 25.0 (now with RECOAlCaEle)
    - fixed 140.0 name map
    - updated workflow directory names in anticipation of merging RECO and miniAOD
- updated validate.C to correctly plot PF candidates for heavy ions files


slava77:
@degano @smuzaffar 
please pick up the changes or comment if you see any issues

PR #506, opened 2015-07-06
add Calibration/Hotline to AlCa category

slava77:
https://github.com/cms-sw/cmssw/pull/9802 was merged already without a new category for the package.


slava77:
@smuzaffar @degano 
please check

PR #509, opened 2015-07-20
new workflows/names in the map file for fwlite-based comparisons

slava77:
will be missing 1001 otherwise


slava77:
superseded by #520

PR #520, opened 2015-07-30
fix comparison map 

slava77:
- fix badly formatted TTbarwf25p0 
- add ZMMPUwf25206p0
- new workflow names for 4.77 and 1001.0
... wildcard-based solution is still pending


slava77:
@smuzaffar @degano 
please check

PR #530, opened 2015-08-19
add DQM/PixelLumi to dqm category

slava77:


davidlt:
I have committed exactly the same change.

PR #538, opened 2015-09-07
add RecoBTag/CTagging to reconstruction category

slava77:
to go with https://github.com/cms-sw/cmssw/pull/11122


slava77:
@smuzaffar @degano please check and merge if it's OK

PR #543, opened 2015-10-05
Add more HGC packages to be watched

lgray:
Add in ForwardDetId, HGCalSimData, HGCRecHit, HGCDigi.

Add appropriate watchers as well.

@bsunanda @pfs @vandreev11

@sethzenz @kpedro88 You can add yourselves if you want by forking cms-bot and editing the watchers.yaml file as seen below.

PR #544, opened 2015-10-05
add self to watch HGC packages

sethzenz:
Adding myself to watch HGC packages as suggested by @lgray in #543


lgray:
@sethzenz please close this, it is included in #546

PR #545, opened 2015-10-06
Mount cvmfs for docker image as it's needed for some relval workflows.

degano:


davidlt:
To be precise, GEN (LHE files) are stored on CMVFS and are not managed by CMSDT. This is because those LHE files are specific for each campaign IIUC. That means, each time they want something new, they generate new LHE files and get them available on CVMFS. Thus GEN does not support running production on non-CVMFS machines.

PR #557, opened 2015-10-21
Fix typo in Cmsdist tests.

degano:


smuzaffar:
I think you should also get CMSDIST_BRANCH from the PR (something like obj["base"]["ref"]
and for ARCH, you can use config.map to find out the archs available for CMSDIST_BRANCH and if multple archs are matched then use the one with PROD_ARCH=1

PR #560, opened 2015-11-03
Comparisons setup updates (maps; more variables to plot)

slava77:
in the map definition:
- add run2 workflows; 
-update some older wf directory names

in the validate.C comparisons
- added photon and electron shower shapes used in new regressions; 
- add method3 (eaux) energy for HBHE hits; 
- added muonShower variables; plot gedPhotonsTmp, gedGsfElectronsTmp, ecalDrivenGsfElectrons and mvaElectrons (relevant for HI); 
- fix few leftover hardcoded RECO process (only in uncleanedHybridSuperClusters); 
- drop plots of Ref().isAvailable() (it's crashing occasionally and was never really useful); 
- new b-tags combinedMVAV2BJetTags, pfCombinedMVAV2BJetTags, pfCombinedCvsLJetTags and pfCombinedCvsBJetTags

@smuzaffar @degano  (in case you are not watching the repo by default)


smuzaffar:
@slava77 , any idea how much more time it would take to run this now. Comparison is already taking one hour for the existing workflows

slava77:
@smuzaffar 
all variables added here are quick to plot.
The time-consuming plots are in particleFlow_ branch and (less so) in generalTracks.
I will try to trim these down (not in this PR though).

PR #565, opened 2015-11-17
Add SLHCDEV specific workflow list and add 108xx tests to it

mark-grimes:
This splits off the list of SLHCDEV test workflows and SLHC workflows.  The SLHCDEV workflows are generally quite different to the SLHC ones.  In SLHC the workflow numbering is fixed for stability, whereas in SLHCDEV old obsolete numbers are reused for new test geometries.

In addition, two 108xx (Extended2023Muondev) tests have been added to the SLHCDEV tests.  Extended2023Muondev is the test geometry for the new Phase2 geometry that will eventually become the Phase2 standard in 8_0_X, so it is important to have test coverage on this.  Even though SLHC/SLHCDEV are on the way out tracker are using it as a stepping stone for migration to 8_0_X.

**Note that I have done barely any testing on this** other than checking with a dummy script that the correct case path is followed.  I have no idea how to test the modified script locally, if there is a way let me know.

@fratnikov


mark-grimes:
please test

Are there any tests for cms-bot?  Or would the concept cause github to be destroyed in a meta-recursive implosion?

degano:
@mark-grimes ah! no there are no tests for this repository, I guess we'll just have to trust you :D

PR #576, opened 2015-11-25
update comparison matrix map for 75X and 71X

slava77:
update 75X name of SingleMu13Pt1wf1306p0 PR.
Noticed by the bot in https://github.com/cms-sw/cmssw/pull/12505#issuecomment-159538885

also updated 71X cases for TTbarPUwf25202p0 and TTbarPUwf50202p0

A more meaningful solution is to use wildcards, but I didn't get there yet after ~3 year history of updating map files.


slava77:
apparently, the update is relevant for 76X as well

@smuzaffar @degano (in case you are not watching this repository otherwise)

PR #588, opened 2015-12-02
Also check toolfiles in lcg dir

degano:


smuzaffar:
there are tool files in cms directory too

PR #597, opened 2016-01-15
Send log errors and msgs to elasticsearch code added

iahmad-khan:


runPyRelValThread.py
```diff
@@ -197,6 +198,7 @@ def parseLog(self):
         logFile = logData[wf]['steps'][step]
         json_cache = os.path.dirname(logFile)+"/logcache_"+str(step)+".json"
         if (not os.path.exists(json_cache)) or (os.path.getmtime(logFile)>os.path.getmtime(json_cache)):
+          log_payload(logFile)
```
smuzaffar: add try catch block here so that runPyRelValThread does not fail due to error in new code

smuzaffar:
rename process_payload.py to process_relval_log.py

es_relval_log.py
```diff
@@ -0,0 +1,60 @@
...
+import dateutil.parser
+from socket import gethostname
+from elastic_send import send_payload
+
+def log_payload(logFile):
+  payload = {}
+  architecture = logFile.split('/')[4]
+  release = logFile.split('/')[8]
+  workflow = logFile.split('/')[10].split('_')[0]
+  step = logFile.split('/')[11].split('_')[0]
```
smuzaffar: please split once

PR #601, opened 2016-01-26
Do not use twice more threads than available virtual cores

davidlt:
After running several CMSSW compilation tests on ppc64le and x86_64 with
X and 2 * X threads (-j), where X is number of available virtual cores, I
have not detected any significant difference in elapsed time.

This is expected due to the nature of our heavy translation units and
a long debug information optimization step.

Thus moving to X + 1 should make sure we are not swapping while
compiling CMSSW while still maintaining same elapsed time.

Signed-off-by: David Abdurachmanov <David.Abdurachmanov@cern.ch>


davidlt:
In addition to that we don't want to run -j 320 on 160 virtual core machine as on high-core machines we might end up having less memory per core. E.g. new generation x86_64 machines we have are 72 virtual core with 64GB of DDR4 memory.

PR #607, opened 2016-01-29
remove 13600 and 13661 for IB and PR tests of SLHC releases

smuzaffar:


smuzaffar:
This was requested in cms-sw/cmssw#13127

PR #615, opened 2016-02-09
elasticsearch log parser updated

iahmad-khan:


smuzaffar:
remove these extra commits

PR #622, opened 2016-02-19
timestamp info changed

iahmad-khan:


iahmad-khan:
timestamp info changed

iahmad-khan:
Time stamp variable name changed for kibana to parse

PR #623, opened 2016-02-19
timestamp info changed to datetime format

iahmad-khan:


iahmad-khan:
The index mapping date format require the variable to be stored as date-time object

PR #627, opened 2016-02-21
[RFC] cvmfsInstall.sh: add PRoot/QEMU environment support

davidlt:
aarch64 and ppc64le should use QEMU/PRoot + custom rootfs setup for
installing releases.

Signed-off-by: David Abdurachmanov <david.abdurachmanov@gmail.com>

### DIFF
--- a/cvmfsInstall.sh
+++ b/cvmfsInstall.sh
@@ -4,6 +4,7 @@ CMS_WEEK=$2
 RELEASE_NAME=$3
 export BASEDIR=/cvmfs/cms-ib.cern.ch
 export BASEDESTDIR=/cvmfs/cms-ib.cern.ch
+export PROOTDIR=/build/cmsbuild/proot
 export THISDIR=`pwd`
 export LANG=C
 # The disk where cvmfs is mounted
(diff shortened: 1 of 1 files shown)


davidlt:
Do not merge, this is RFC.

cvmfsInstall.sh
```diff
@@ -57,12 +58,23 @@ for dir in $(find $BASEDESTDIR/* -maxdepth 0 -type d | grep -G "20[0-9][0-9]-[0-
...
-  else
-    eval $@
-  fi
+  case "$SCRAM_ARCH" in
+    slc7_amd64_* )
+      ARGS="cd $THISDIR; $@"
+      docker run -t -e THISDIR=${THISDIR} -e WORKDIR=${WORKDIR} -e SCRAM_ARCH=${SCRAM_ARCH} -e x=${x} -v ${WORKDIR}:${WORKDIR} -v ${THISDIR}:${THISDIR} -u $(whoami) cmssw/slc7-installer sh -c "$ARGS"
+      ;;
+    slc7_aarch64_* )
+      ARGS="export THISDIR=${THISDIR}; export WORKDIR=${WORKDIR}; export SCRAM_ARCH=${SCRAM_ARCH}; export x=${x}; cd ${THISDIR}; $@"
```
davidlt: Actually it would be smarter to do `export THISDIR='${THISDIR}'`, just in case there is a whitespace in these variables, but that should be unlikely.

cvmfsInstall.sh
```diff
@@ -4,6 +4,7 @@ CMS_WEEK=$2
 RELEASE_NAME=$3
 export BASEDIR=/cvmfs/cms-ib.cern.ch
 export BASEDESTDIR=/cvmfs/cms-ib.cern.ch
+export PROOTDIR=/build/cmsbuild/proot
```
davidlt: I looked and Jenkins job parameters and environment variables, I couldn't find a good reference. I guess, this has to be hardcoded somewhere. We could leave it here and discussing the actual path we want to use, or move it to machine specific environment variable in Jenkins. I probably prefer to keep it in this script.

PR #636, opened 2016-02-25
add 2017 tenMuE and ttbar to comparisons matrix

slava77:
- added resim names for 2017 scenarios
- added 10039.0,10060.0 to the tested workflows
    - in 76X these don't exist and are simply skipped , e.g. https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-13392/11543/runTheMatrix-results/matrixTests.log
    - in 80X and 81X they run OK, e.g. https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-13452/11542/runTheMatrix-results/

@boudoul @hengne
We can update the list or add some of the workflows to the short matrix.


smuzaffar:
@slava77 , is this safe to merge or do you need anything else to be updated?

slava77:
@smuzaffar 
I think it's safe to merge

smuzaffar:
merge

slava77:
@smuzaffar 
is there a build step after this or is jenkins going to pick up the merged code immediately?

smuzaffar:
jenkins it going to pick up immediately

slava77:
Now I realize that I didn't check how the baseline reference is built.

If I add new matrix in  run-ib-comparison-baseline, I probably need to wait for another IB baseline build to actually get everything setup.
Is this correct?

smuzaffar:
yes exactly. New workflows e.g. 10039.0,10060.0 will only be available in next Baseline

slava77:
Looking at https://cmssdt.cern.ch/jenkins/job/ib-any-integration/11576/parameters/
I don't see the new workflows.
Are they entered in a different list?

I guess I don't know all the internals.

slava77:
To clarify what I did, I just did a git grep for 50202.0 (one of the extra wflows) and appended the new workflow to the list there.

smuzaffar:
You are right, we have duplication here. There is 
- one list which is in run-ib-comparison-baseline script. So next Baseline will use it
- and then there is a list in jenkins job ib-any-integration (I totally forgot about this).

I will update the jenkins list to have the next workflows too. I will try to make it configuration via cms-bot so that next time we only update it in one place.

slava77:
On 2/26/16 6:34 AM, Malik Shahzad Muzaffar wrote:
> You are right, we have duplication here. There is
> 
>   * one list which is in run-ib-comparison-baseline script. So next
>     Baseline will use it
>   * and then there is a list in jenkins job ib-any-integration (I
>     totally forgot about this).
> 
> I will update the jenkins list to have the next workflows too. I will
> try to make it configuration via cms-bot so that next time we only
> update it in one place.

Thank you.

I see that jenkins already made baseline for 02-26-1100.
So, the next up is 26-2300 which may or may not take time to build.
No rush, just taking a note.

slava77:
@smuzaffar 
How does the MATRIX_EXTRAS propagate into the schedule-pr project?
Is the default in the job config overriding what you have implemented now ?

The config now has the old values:
jenkins_scheduleprtetst

slava77:
I remove the default in jenkins. Now it looks like the bot will use the ones from the cms-bot repo code.
The last rebuild that was started picked up the defaults now from the cms-bot itself https://cmssdt.cern.ch/jenkins/job/ib-any-integration/11600/console

hengne:
@slava77 Hi Slava, I am cleaning up the phase1 relvals and the workflow numbers are changed. I will send a PR later this evening, and will let you know the new workflow numbers to be used. 

best, Hengne

smuzaffar:
@slava77, thanks for cleaning up ib-schedule-pr-test job. I did not noticed that this job was also provide workflow list to ib-any-integration job. 
You did the right thing, if MATRIX_EXTRAS is not set via jenkins job then the default values from cm-bot will be used.

PR #637, opened 2016-02-26
Check for status change of running builds to stale-deleted added

iahmad-khan:


parse_jenkins_builds.py
```diff
@@ -31,6 +90,11 @@
           os.system('touch "' + flagFile + '"')
         else:
           payload['job_status'] = 'Running'
+          running_current.append(id)
         send_payload(index,document,id,json.dumps(payload))
       except Exception as e:
         print "Xml parsing error" , e
+for build in running_builds_elastic:
+  if build not in running_current:
+    send_payload(index,document,build,'{"job_status":"unkown-stale"}')
```
smuzaffar: Please change the job_status to failure here

smuzaffar:
once PR is requested then please give us sometime to go through it

parse_jenkins_builds.py
```diff
@@ -31,6 +90,11 @@
           os.system('touch "' + flagFile + '"')
         else:
           payload['job_status'] = 'Running'
+          running_current.append(id)
         send_payload(index,document,id,json.dumps(payload))
       except Exception as e:
         print "Xml parsing error" , e
+for build in running_builds_elastic:
+  if build not in running_current:
```
smuzaffar: logic is not correct here. Instead of using running_current use all local jobs. If a running_builds_elastic job is seen locally then this means for that status has been updated. So we do not want to mark them as unkown-stale/failed

PR #643, opened 2016-03-11
Remove permissions for mark-grimes

mark-grimes:
I no longer work on CMS so shouldn't have permission to merge/request builds/etcetera.  I haven't tested this because I don't know how, so check I haven't made a dumb typo.

So long, and thanks for all the memories.


smuzaffar:
looks good.
Thanks for taking care of SLHC releases

PR #660, opened 2016-04-12
DataFormats/TotemRPDigi -> TotemDigi to consolidate common types for other Totem detectors

slava77:


smuzaffar:
+externals

smuzaffar:
merge

slava77:
@mdhildreth @civanch these Digis will need to move/be added to sim category at some point.
There is no simulation with TotemDigi right now, but it's coming

PR #676, opened 2016-05-17
adds parsing for iwyu log files

iahmad-khan:


smuzaffar:
@iahmad-khan , can you please create jenkins jobs to run it?

PR #678, opened 2016-05-18
updates iwyu log to html pages

iahmad-khan:


parse_iwyu_logs.py
```diff
@@ -15,33 +15,37 @@
...
-      print '<tr><td bgcolor="#00FFFF"><h2>'+l[70:].replace('lines','includes')+'</h2>'
-      while line.startswith('-'):
-        line=line.replace('<','')
-        line=line.replace('>','')
-        line=line.replace('"','')
-        line=line.replace('- #include ','')
-        print '<br/>'+line.split('//')[0]
+    line=line.rstrip()
+    if len(line):
+      print '<tr><td bgcolor="#00FFFF"><h2>'+'/'.join(l.split('/')[-4:])+'</h2>'
```
smuzaffar: wrong assumption here. File could be in sub-directory e.g src/Subsystem/Package/src/subdir/file.cc
[-4:] will return /Package/src/subdir/file.cc which is wrong

iahmad-khan: ok , i ll fix it.

PR #691, opened 2016-05-27
updates iwyu related files for ES

iahmad-khan:


smuzaffar:
does not look like a valid PR

PR #710, opened 2016-06-13
Comparisons update: add CTPPS monitoring and update map file

slava77:
CTPPS monitoring is needed following integration of 
https://github.com/cms-sw/cmssw/pull/14605

The map file update includes addition of 2016 and 2023 workflows and should also resolve the issue reported in recent PR tests, e.g. in https://github.com/cms-sw/cmssw/pull/14853#issuecomment-225594349

without an update we are currently blind to comparisons of run2 data


slava77:
@smuzaffar @degano 
please let me know if there are any issues with this PR

PR #717, opened 2016-06-22
Ignore missing files in the python check

ericvaandering:


ericvaandering:
@smuzaffar this is blocking some tests of ours. Can you push this in ASAP?

PR #720, opened 2016-06-28
2017 WFs now in limited, not needed as extra tests

kpedro88:
NOTE: this should *not* be merged until https://github.com/cms-sw/cmssw/pull/14977 is merged *and* in an IB.

attn: @slava77


kpedro88:
@davidlange6 @smuzaffar cms-sw/cmssw#14977 is in the IB now, so this can be merged.

PR #733, opened 2016-08-23
remove obsolete hcal modules

deguio:
this update follows PR:
https://github.com/cms-sw/cmssw/pull/15374


smuzaffar:
@deguio , we can not remove these package as they are part of other releases e.g. 53X, 7X etc.

deguio:
ok, I see. no problem then. we can leave them in the categories file.
F.

PR #766, opened 2016-10-29
Add fast timing digi and digitization simulation

lgray:
These packages are proposed in:
https://github.com/cms-sw/cmssw/pull/16387


lgray:
@davidlange6 OK - updated here too.

PR #771, opened 2016-11-07
adds support for new es cluster to use in prarallel to old

iahmad-khan:


iahmad-khan:
@smuzaffar should we merge it? keys deployed already

smuzaffar:
wait

PR #787, opened 2016-12-06
Add fast-timing rechits and local reconstruction packages

lgray:
This is coming soon in a PR.


categories.py
```diff
@@ -245,6 +246,9 @@
    "RecoLocalCalo/HcalRecProducers",
    "RecoLocalCalo/HGCalRecAlgos",
    "RecoLocalCalo/HGCalRecProducers",
+   "RecoLocalFastTime/Configuration",
+   "RecoLocalFastTime/FTLCommonAlgos",
+   "RecoLocalFastTime/FTLRecProducers",
```
slava77: Do these have to have "Common" and "Rec" in the name?
What else do we expect in the RecoLocal for FTL?

lgray:
As discussed by skype with @slava77, not going to change it.

lgray:
This is needed for: https://github.com/cms-sw/cmssw/pull/16901

PR #795, opened 2016-12-16
validateJR plots: fix plotting for ref with a different process name; add plots for boosted taus …

slava77:
Somewhat minor updates in validateJR

- plotting for ref with a different process name: it was already possible to plot for two different processes but selection string was not updated, this PR fixes it. This mode is not used in standard comparisons yet

- miniaod plots: added monitoring of slimmedTausBoosted, just a copy of what's done for regular taus.


slava77:
superseded by #797

PR #806, opened 2017-01-31
add upgrade category for PRs w/ list of packages

kpedro88:
This PR adds a new category "upgrade" for CMSSW pull requests with a list of included packages.

The list primarily includes Phase2-specific packages for tracker, HGCal, fast timing, + a few for EB. I also included the GEM/ME0 packages even though they aren't strictly Phase2. A few other packages were added due to frequent phase2-related interactions:
    "Configuration/Geometry",
    "Configuration/PyReleaseValidation",
    "Geometry/CMSCommonData".

Please send any comments.


smuzaffar:
@slava77 , @davidlange6 if no objections then I can merge it
Note that the existing PRs (where these packages are involved) shall only get the new category once someone commented on them.

slava77:
looks OK to me, although I'm not sure about the more generic packages which probably already have multiple signatures

kpedro88:
I tried to add as few of those as possible, but the ones I included are probably the most frequent causes of upgrade workflows breaking...

davidlange6:
so two issues:
 - we agreed "dual use" packages would not get an upgrade sig by default. (but rather one added by hand when relevant
 - i had asked only sim/reco/l1. if you want also dqm areas we'll have to do more inquiries. This is particularly asymmetric compared to the organization of the rest of the stack, however. So not sure that is a good way to organize.

kpedro88:
* I know, but I was hoping I could get away with just those few. They have a high concentration of upgrade-related code.
* I just grabbed all the phase2-specific packages for the list. I would personally be quite happy to have no responsibility for validation code, but it is a rather important area for upgrade studies.

If the final ruling is to remove those things from the list, I can live with that.

kpedro88:
@davidlange6 was the initial list ultimately acceptable? (We can always revise it later if necessary.)

davidlange6:
they aren't - I'll update slightly to be organizationally consistent

kpedro88:
while you're at it, can you add ` L1Trigger/TrackTrigger`? I missed that one.

davidlange6:
done - the one dual use package is Configuration/PyReleaseValidation - it may become too annoying but for now its there.

kpedro88:
Configuration/Geometry and Geometry/CMSCommonData are also still on there. Is that okay? I would prefer to keep them due to numerous instances of people screwing up the Phase2 geometries

PR #807, opened 2017-01-31
add HLT muon packages to the "hlt" category

fwyzard:


categories.py
```diff
@@ -625,20 +625,33 @@
...
+   "HLTrigger/special",
+   "RecoEgamma/EgammaHLTAlgos",
    "RecoEgamma/EgammaHLTProducers",
-   "HLTrigger/JSONMonitoring",
+   "RecoMuon/L2MuonIsolationProducer",
+   "RecoMuon/L2MuonProducer",
+   "RecoMuon/L2MuonSeedGenerator",
+   "RecoMuon/L3MuonIsolationProducer",
+   "RecoMuon/L3MuonProducer",
+   "RecoMuon/L3TrackFinder",
```
smuzaffar: @fwyzard , should these RecoMuon/L2* packages belong to ```hlt```?

fwyzard: I think so, RecoMuon/L2* and RecoMuon/L3* are only used for HLT as far as I know.

@sarafiorendi , @BenjaminRS , can you confirm ?

BenjaminRS: Yes, the code contained in RecoMuon/L2* and RecoMuon/L3* are for HLT purposes.

davidlange6: @slava77 - fyi - but I agree these are only used in HLT

smuzaffar: if no objections from @slava77 then I am happy to merge this

slava77: these just add the signatures to existing packages.
So, go ahead.

PR #808, opened 2017-02-02
comparisons: switch fw-lite comparisons to regexp-based map; add deepCSV plots

slava77:
- introduce matrix_RE.txt which uses regular expression for path names (limited to the shell path expansion capabilities)
- updated matrix_70X.txt to new names .. this is updated only in case matrix_RE.txt switch fails for some reason
- added a few plots for new deepCSV tagger in validateJR
- update  comparisons/validateJR.sh  to handle matrix_RE.txt format

this is supposed to also fix issues with missing (JR) comparisons that appeared sing Jan 30/31.


slava77:
@smuzaffar 

this transition works for me locally,
but the implementation of the alternative-comparisons driver is different.
I think I updated the right places in cms-bot
Still, there is a small chance that there will need to be a followup

smuzaffar:
@slava77 , before merging it, should we re-run comparison for any PR with your changes to see if it works as expected.

slava77:
the only failure modes I can think of are from missing the workflow inputs completely.
The actual set of plots in each workflow comparison would not matter because that part was not touched.

It would be good to run this before merging on any selected PR, but it should be done with all relevant setup as inside jenkins tests. IIRC, this means that it will actually need to be merged because there is no testbed.

smuzaffar:
for comparison we have jenkins job option to use a different cms-bot. I can just re-start an existing comparison job to use your branch and see how it goes.

slava77:
good. please post here where/when it's done.
Thank you.

smuzaffar:
Any PR number you would prefer, I should I just pick a random one?

slava77:
a random one is good enough

smuzaffar:
it is running here https://cmssdt.cern.ch/jenkins/job/compare-root-files-short-matrix/18126/console

smuzaffar:
@slava77 , comparison with your changes are now available here
https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_9_0_X_2017-02-01-1200+17378/18126/

slava77:
The outputs look good: the validateJR directories are in the right place with familiar names and the logs show that they ran fine.
So, it looks like we can merge it.

PR #824, opened 2017-03-23
changes for using Cern ElasticSearch

iahmad-khan:


smuzaffar:
we should use the same template/mapping for both CERN IT based ES and our ES. We can have a jenkins job to push with different mapping name for CERN It

PR #832, opened 2017-04-13
reco comparisons: add CTPPS pixel digis and muon puppi isolation; recover packedCandidates physics level plots

slava77:
- packedCandidates physics level plots are actually working (they were commented out/broken in early 7X version) ==> uncommented now
- add CTPPS pixel digi plots

### DIFF
--- a/comparisons/validate.C
+++ b/comparisons/validate.C
@@ -14,6 +14,7 @@
 #include "TCut.h"
 #include <cmath>
 
+
 bool detailled = true;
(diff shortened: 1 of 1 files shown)


slava77:
packed candidate plots area apparently not that good for all workflows (despite working in the ones that lead me to believe the plots work).
I may reopen after investigation.

slava77:
I have reopened this after fixing issues

@smuzaffar  please check and merge some time soon
Thank you.

PR #833, opened 2017-04-18
Update categories.py

mrodozov:
adding mrodozov as external manager for CMSSW software


smuzaffar:
+externals
@mrodozov is new to our group CMSSDT and he will be taking care of  building externals.

PR #836, opened 2017-04-20
add all subdirectories of the package HLTriggerOffline

mtosi:


smuzaffar:
@mtosi , if you are interested in all packages of HLTriggerOffline then better to use ```- HLTriggerOffline/``` this will allow you to watch any new HLTriggerOffline packages too

mtosi:
thanks @smuzaffar !
I've just updated the file

PR #837, opened 2017-04-22
comparisons updates: add pfDeepCMVA tag and muon matches

slava77:
- improve packedCand comparisons
- add basic plots of muon matches 
- new products pfDeepCMVAJetTags


slava77:
ping

this one is not particularly urgent, but it would still be nice to have some time soon.

PR #845, opened 2017-05-11
Updating run-material-budget script 

mrodozov:
Changing the script to work with either the .C script or the .py regardless if the .py script will be backported to older releases


smuzaffar:
does not look good, there are merge conflicts

PR #863, opened 2017-06-14
Adding Kaori and Virginia as L2 for DQM

vanbesien:
Adding current DQM conveners Kaori Maeshima and Virginia Azzolini as L2 for DQM
dmitrijus remains as main responsible for reviewing and signing PRs
vanbesien will/should be removed end of September


smuzaffar:
FYI @dmitrijus , @davidlange6

PR #883, opened 2017-08-02
Add swertz to watchers.yml

swertz:


swertz:
Thanks for the quick merge!

PR #903, opened 2017-09-27
add Giovanni P to test trigger

slava77:


slava77:
@gpetruc 
to make it easier for you

PR #906, opened 2017-10-09
Lizard .txt to .html conversion script

gudrutis:
Created a python script which converts simple lizard.py output to separate .html documents with tables.
They can be sorted, searched and have click-able hyper-links to source code at given repository.


smuzaffar:
@gudrutis , please create a separate lizard directory for this

PR #911, opened 2017-10-24
Comparisons: urgent fix to patJet plots; add MuonSimInfo plots

slava77:
Update comparisons
- fwlite-based validate.C:
    - fix plots of jets fractions which broke miniAOD reco validation since merging of https://github.com/cms-sw/cmssw/pull/20638
    - add plots of MuonSimInfo , introduced with https://github.com/cms-sw/cmssw/pull/20773
- also added a minor update to the map file to cover premix workflows (used in private tests now)


slava77:
@smuzaffar @davidlange6 
please check.
We are currently missing all reco miniAOD comparisons due to the exception thrown in the comparisons script which is not detected as a problem by the driver script.

davidlange6:
i think the script is indeed configured not to stop if a subscript fails..

can you give an example problem comparison?

slava77:
most of the miniAOD comparisons will show the problem
e.g. https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_9_4_X_2017-10-24-1100+20997/23273/validateJR/all_mini_OldVSNew_RunSinglePh2017B136p788/all_mini_OldVSNew_RunSinglePh2017B136p788.log
ends with 
```
Exception Message:
This PAT jet was not made from a JPTJet nor from PFJet.
```
So, for the last two weeks we were going blind in miniAOD comparisons.

davidlange6:
so what is the desired behavior? Made we can add a message to the comparison output in case a script fails  - is that what you have in mind?

I presume you'll get the exception fixed..

slava77:
On 10/25/17 9:18 AM, David Lange wrote:
> so what is the desired behavior? Made we can add a message to the
> comparison output in case a script fails - is that what you have in mind?
> 

the desired behavior is that the script should proceed to completion.
It would be nice eventually to detect the problem at QA level of the
comparison results
if it does not for some reason.

> I presume you'll get the exception fixed..

it is "fixed" for this specific case in this PR with try-catch.
I will add a nicer solution once Sal sorts out what is actually desired
in this part
https://github.com/cms-sw/cmssw/pull/21006


> 
>> On Oct 25, 2017, at 9:00 AM, Slava Krutelyov

davidlange6:
> On Oct 25, 2017, at 9:30 AM, Slava Krutelyov <notifications@github.com> wrote:
> 
> On 10/25/17 9:18 AM, David Lange wrote:
> > so what is the desired behavior? Made we can add a message to the
> > comparison output in case a script fails - is that what you have in mind?
> > 
> 
> the desired behavior is that the script should proceed to completion.
> It would be nice eventually to detect the problem at QA level of the
> comparison results
> if it does not for some reason.
> 


"the script" means validate.C in the end? so it seems you have now fixed up that script. 

It looks like validateJR.sh can also be simply changed to ignore non-0 exit codes for validate.C (except it looks like 0 is the only possible exit code, so no need). 

Propagating these up is not simple..maybe your try-catch block should add a file to the output directory that indicates a comparison failed. I think I can pick that up in the logRootQA script.

> > I presume you'll get the exception fixed..
> 
> it is "fixed" for this specific case in this PR with try-catch.
> I will add a nicer solution once Sal sorts out what is actually desired
> in this part
> https://github.com/cms-sw/cmssw/pull/21006
> 
> 
> > 
> >> On Oct 25, 2017, at 9:00 AM, Slava Krutelyov

slava77:
On 10/25/17 9:52 AM, David Lange wrote:
> 
> 
>> On Oct 25, 2017, at 9:30 AM, Slava Krutelyov
> <notifications@github.com> wrote:
>>
>> On 10/25/17 9:18 AM, David Lange wrote:
>> > so what is the desired behavior? Made we can add a message to the
>> > comparison output in case a script fails - is that what you have in
> mind?
>> >
>>
>> the desired behavior is that the script should proceed to completion.
>> It would be nice eventually to detect the problem at QA level of the
>> comparison results
>> if it does not for some reason.
>>
> 
> 
> "the script" means validate.C in the end? so it seems you have now fixed
> up that script.

yes, it is fixed.
I was just asking to get this merged, not to do some extra work first,
so that we can get the miniAOD comparisons running again.


> 
> It looks like validateJR.sh can also be simply changed to ignore non-0
> exit codes for validate.C (except it looks like 0 is the only possible
> exit code, so no need).
> 
> Propagating these up is not simple..maybe your try-catch block should
> add a file to the output directory that indicates a comparison failed. I
> think I can pick that up in the logRootQA script.

The solution would have to be the opposite: normal behavior should
produce a file
at the end to indicate successful completion.

>

slava77:
@davidlange6 @smuzaffar 
please let me know if there is a problem with this PR that should be addressed that warrants still keeping miniAOD reco comparisons broken without a fix in this PR.

PR #919, opened 2017-11-22
Add new packages for online meta-data

mommsen:
Add new packages for online meta-data format and unpacker.


slava77:
@smuzaffar 

me and  @perrotta discussed this and it is OK to have the new package(s) assigned to reco, in addition to the daq category.
So, this PR is OK for us.

PR #925, opened 2017-12-04
add PhysicsTools/TensorFlow to reco/at and RecoBTag/DeepFlavour to reco

slava77:
new packages added to categories.py

add PhysicsTools/TensorFlow to reco/at and RecoBTag/DeepFlavour to reco


slava77:
this is needed for #19893

PR #933, opened 2017-12-20
Update gh-teams.py (generators-l2)

perrozzi:


smuzaffar:
invitation has been sent to @perrozzi and @efeyazgan to join the team.

PR #949, opened 2018-02-12
reco comparisons updates: add plots of user data for leptons and photons

slava77:
add plots of user data (ints, floats, isolations, IDs) for leptons and photons


slava77:
@smuzaffar 
please let me know if you see any issues with this PR.

PR #961, opened 2018-03-13
remove jlagram from watching

jlagram:
Removing myself and old contributors having left: jlagram, cnuttens, lbeck, sdevissc


smuzaffar:
thanks @jlagram
@cnuttens , @lbeck , @sdevissc any objections?

PR #978, opened 2018-04-25
Add a 2018 premixing workflow to PR tests

makortel:
The workflow was added in https://github.com/cms-sw/cmssw/pull/22956. Resolves https://github.com/cms-sw/cmssw/issues/22586.


kpedro88:
Usually when I want to add something as a PR test, I just update the short matrix:
https://github.com/cms-sw/cmssw/blob/master/Configuration/PyReleaseValidation/scripts/runTheMatrix.py#L50

smuzaffar:
@makortel , yes its better to update the short martix so that new workflow only run for that release cycle.
Although, runTheMatrix just ignore the workflow which it does not know, so this is not going to break earlier IBs PR tests.

makortel:
Let me then ask why the workflows defined here are not in short matrix?

Also the short matrix does not have any pileup workflows (the two such workflows running in PR tests are defined here). Is this a deliberate choice or more of an "accident"?

(I'm personally not against of adding this specific workflow to short matrix, just want to understand first if there is a desire to not add pileup workflows there)

@fabiocos @davidlange6 @slava77

slava77:
> Let me then ask why the workflows defined here are not in short matrix?

I don't fully recall the historical origins of this.
Perhaps when these extras were added in the jenkins bot setup:
- it was considered as some initial testing, to be propagated "later" to the release
- there was lack of feeling of responsibility for editing Configuration/PyReleaseValidation/scripts/runTheMatrix.py file
- the additional time and memory needed to run the PU workflows seemed too hig

smuzaffar:
I also do not know the history of why these workflows were kept in the bot (may be we did not want to have a long list workflows in short matrix so that developers can run/tests before the PR submission).

fabiocos:
In case we decide to move from extra additions to the regular cmssw matrix, cms-sw/cmssw#23073 provides the code

makortel:
So how do we decide whether the solution is this PR or cms-sw/cmssw#23073? (I'd like to get the premixing workflow included rather sooner than later)

I have vague recollection of complaints on `runTheMatrix.py -l limited` taking already now/then long to run.

(btw, should we replace the 2017 PU with 2018 PU?)

smuzaffar:
Lets get this PR merged. If needed we can add the workflow in the short matrix

PR #982, opened 2018-05-07
updated the categories_map

chenxvan:
updated the categories_map.py for the pixel dqm


smuzaffar:
-externals
@chenxvan , please do not remove packages from the category. Although these packages are not part of cmssw master but they do exist in earlier release cycles, so we need to keep them in the mapping.

chenxvan:
@smuzaffar Hi, I just added the old packages back

PR #992, opened 2018-06-04
Update openstack image names for CC7 and SLC6

amaltaro:
The old March images are no longer available.
@smuzaffar please review and merge it, apparently this is why jenkins "openstack-create-dmwm-vms" jobs fail.


smuzaffar:
thanks @amaltaro

PR #998, opened 2018-06-19
Add the new CondFormats/SiPixelTransient package, and assign both db and reco as owners

perrotta:
Add the new CondFormats/SiPixelTransient package, and assign both db and reco as owners
The new package is introduced by #23571


smuzaffar:
+externals

PR #1001, opened 2018-06-29
Auto-discover CC7 and SLC6 images available for DMWM

amaltaro:
@smuzaffar the openstack-create job is failing again, and it turns out the previous image is no longer available. 
This script should be able to discover what's the image name available (assuming only one CC7/SLC6 match)


smuzaffar:
thanks @amaltaro

PR #1005, opened 2018-07-06
Update categories.py

dmitrijus:
Replace Broen with Marcel Schneider as DQM L2 convener.


smuzaffar:
+externals
@fabiocos , any objections?

fabiocos:
@jfernan2 I assume this is ok for the DQM group, right? Obviously fine for me

smuzaffar:
as @fabiocos has no objections and @dmitrijus (dqm L2) requested it, so merging it now.

PR #1030, opened 2018-09-13
Update categories.py

fabozzi:
Leaving only current PdmV conveners


smuzaffar:
thanks @fabozzi for the clean up.

PR #1035, opened 2018-09-24
Update watchers.yaml

OlivierBondu:
Removing myself from following SiStrip LocalReco packages (stepping down)


smuzaffar:
@OlivierBondu , thanks for cleanup

PR #1037, opened 2018-10-01
Update categories_map.py

mundim:
Including PPS modules into categories map.


categories_map.py
```diff
@@ -1445,6 +1446,10 @@
     "SimG4CMS/HGCalTestBeam",
     "SimMuon/GEMDigitizer",
     "SimTracker/SiPhase2Digitizer",
+    "SimPPS/Configuration",
```
kpedro88: @mundim why did you add these packages in the upgrade category? The upgrade category is specific to Phase2. I think these packages should just be placed in the general simulation category.

Issue #1009, opened 2018-07-17
Make it easy to find failures from comparison summary


Dr15Jones:
When the Comparison Summary reports a failure, it is very difficult to find the plots or other pieces of information necessary to understand the failure. Can we provide a way to quickly jump to the relevant information about a failure?

Dr15Jones:
@slava77

Dr15Jones:
@davidlange6

kpedro88:
thumbs up for this!

slava77:
In the recent example https://github.com/cms-sw/cmssw/pull/23764#issuecomment-403248007

- DQMHistoTests: Total failures: 25
detailed info together with the summary displayed in the bot message can be found in
https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_10_2_X_2018-07-06-2300+23764/27366/validateJR/logRootQA.log

To get there: follow the links "Comparison is ready" -> "See Comparison Results" -> "validateJR/" -> "logRootQA.log"
then search for the string in the summary.
The details are right above that.
e.g. for 136.731_RunSinglePh2016B+RunSinglePh2016B+HLTDR2_2016+RECODR2_2016reHLT_skimSinglePh_HIPM+HARVESTDR2RelMonComp-136.731.log 
```[107863, 6, 0, 107848, 9, 0] ```
- 6 is the failed results

you can then find these 6 explicitly in https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_10_2_X_2018-07-06-2300+23764/27366/136.731_RunSinglePh2016B+RunSinglePh2016B+HLTDR2_2016+RECODR2_2016reHLT_skimSinglePh_HIPM+HARVESTDR2/
and more specifically navigate to 3 in one particular folder
https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_10_2_X_2018-07-06-2300+23764/27366/136.731_RunSinglePh2016B+RunSinglePh2016B+HLTDR2_2016+RECODR2_2016reHLT_skimSinglePh_HIPM+HARVESTDR2/L1T_L1TStage2CaloLayer2_Isolated-Tau.html

Perhaps the easiest first step is to provide a link to logRootQA.log in the cms-bot comparisons summary.

smuzaffar:
[...]

PR #1063, opened 2018-11-26
Add reco comparison differences to validateJR.html

kpedro88:
Followup changes for #1009 to make a list of reco comparisons with differences, in addition to failed DQM comparisons.

I tested the bash code standalone on a few logs, and it seems to work fine.


smuzaffar:
thanks @kpedro88

PR #1064, opened 2018-11-27
Add "heterogeneous" category, packages and maintainers

fwyzard:
Clean up the categories: sort and uniform indentation.

Add the "heterogeneous" category and initial  maintainers.
Assign the new packages to the "heterogeneous" category, and to the "reconstruction" category where it makes sense.


fwyzard:
> the aesthetic cleaning (nice but which makes a bit tedious the comparison old - new) 

I understand - that's why I made a first commit with only the clean up and without any content changes, and a second with the changes I wanted to propose.

fwyzard:
>  I would like to hear the opinion of @Dr15Jones and of the reconstruction coordinators @slava77 @perrotta , as originally we agreed to have the new initial only package under core at that time.

I would not say that "we agreed" (if you include myself in that "we").

fwyzard:
> The pure cleaning part could enter immediately (I need to make an unrelated update).

#1067 would take care of that part

makortel:
If we decide to proceed this way, let me just raise a question of "what about the packages outside of the list proposed here that would get CUDA code in https://github.com/cms-sw/cmssw/pull/25353".

fwyzard:
Very valid point.
Requiring the new group to sign for any package that contains CUDA code would likely add too many false positives.

fabiocos:
@makortel @fwyzard with cms-sw/cmssw#28537 now this PR needs to be finalized. Any objection on the proposed. @Dr15Jones @smuzaffar @slava77 @perrotta any objection with the proposed new ```heterogeneous``` category and the related package assignment? I am personally fine with it, understanding that this new category is to some extent a special part of the general framework (a bit in the spirit of ```geometry```).

fwyzard:
OK for me

makortel:
I'm fine with the proposal. I do have two thoughts/comments that I wanted to share though (e.g. for longer term).

I'm wondering if the packages `CUDADataFormats/Common` and `HeterogeneousCore/*` would be under the responsibility of the core L2, or under the responsibility of the individuals signing for `heterogeneous` (ok, little practical difference for me...)? For example, the responsibility of `geometry` category is similar to L3 within core L2.

My feeling of the `heterogeneous` category is that the category is mostly about technology: core components for non-traditional-x86 computing. Currently we don't have categories based on technologies (like ML, vectorization, etc).

makortel:
I also want to note that the list of `CUDADataFormats` sub-packages is incomplete wrt. https://github.com/cms-sw/cmssw/pull/27983, but obviously the additional packages can be added later when needed.

fabiocos:
@makortel in my view this proposal practically allows to delegate part of the ```core``` job to someone (in the spirit of a L3 sub-task within the L2 area, like the ```geometry``` example), with  the possibility to take responsibility to review and allow integration of a part of code with some distinctive characteristics (in this case the underlying technology) instead of calling directly the L2 into the game every time. As you see, now @Dr15Jones is also part of the ```geometry``` area, although he mostly delegates work on that to others. We could manage ```heterogeneous``` in the same way, actually given your new appointment in practice this will become already the case. 
In this case the reason of the specialization of a sub-part is a specific technology, although its implication on the interface is such that probably justify this arrangement (adding a new deep NN package seems less invasive in terms of implications for the framework).

Of course if you or @Dr15Jones have a different preference please clarify. As a matter of fact we need at this point to converge on a decision about the practical management of the newly proposed packages.

makortel:
> Of course if you or @Dr15Jones have a different preference please clarify. As a matter of fact we need at this point to converge on a decision about the practical management of the newly proposed packages.

@Dr15Jones should probably confirm explicitly, but we're fine to start along this proposal, see how it goes, and adjust later if needed.

Dr15Jones:
I'm fine with this as an initial attempt to see how it goes.

fwyzard:
Two pure technical reasons I think it is a good idea to have a dedicated signature for *heterogeneous* packages is to make sure that developers
  - do not reinvent the wheel, and instead use the tools that have been and are being developed;
  - do not accidentally break the computing performance of the code.

fabiocos:
+1

I'll add ```HeterogeneousCore/CUDATest```

fabiocos:
@smuzaffar as soon as it is merged I'll make the final addition needed by cms-sw/cmssw#28537

smuzaffar:
+externals

PR #1066, opened 2018-12-04
Add the package Utilities/PSS to the simulation category

fabiocos:
Addition of the new package Utilities/PPS introduced in cms-sw/cmssw#25355


fabiocos:
@civanch FYI

PR #1085, opened 2019-01-22
reco comparisons: new PPS and MTD plots; add branch presence check for cleaner logs

slava77:
- MTD track extension variables, track tofPID, and new vertex collections
- additional/new PPS local track variables
- some refactoring of copy-pasted values or calls
- add branch presence check for most of the complex types with more than 1-2 plotted members to cleanup the logs
    - e.g. all_mini comparisons logs reduced by more than a factor of 4 from 1.9 MB to 0.4MB (for a case of no differences)

### DIFF
--- a/comparisons/validate.C
+++ b/comparisons/validate.C
@@ -60,12 +60,16 @@ bool stepContainsNU(const TString& s, TString v){
   }
 }
 
-bool checkBranchAND(const TString& b){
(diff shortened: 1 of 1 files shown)


slava77:
@smuzaffar @mrodozov 
please let me know if you see any issues with this PR.

PR #1088, opened 2019-01-28
Update categories_map.py

forthommel:
Adding CTPPSPlugins to the db category, as suggested in https://github.com/cms-sw/cmssw/pull/25777.


forthommel:
Interesting to see this could have been there since https://github.com/cms-sw/cmssw/pull/18359...

PR #1098, opened 2019-03-08
reco comparisons: plots for new products/variables including bParking/lowPt electrons

slava77:
Updates in reco comparisons for the following areas:
- PPS: fix method name in ctppsLocalTrackLiteProducer
- EGM/bParking: plots similar to regular electrons and conversions
- Muons: plots for crossedHadRecHits


slava77:
@mrodozov @smuzaffar 
please check this PR and merge it or let me know if there are any issues.
Thank you.

PR #1108, opened 2019-04-01
reco comparisons: plots for PPS protons

slava77:
- plots for PPS protons
- updated map file to include JetHT workflows (used in local/private tests).


smuzaffar:
+externals

PR #1136, opened 2019-05-10
reco comparisons updates: new plots and updated workflow map

slava77:
New plots:
- electron HEEP iso and conv prob;  photon charged hadron iso and shape 2nd moment
- muon: low-pt MVA
- packed candidates: associated depth value map; caloFraction
- prefiringweight for L1 (non)prefire probability
- another rho variable

Map updates:
- RunEGamma2018C for private tests
- 11624 is now from 2021 (renamed from 2019)
    - this should get rid of the warnings recently appearing in the comparisons run by cms-bot/jenkins

### DIFF
--- a/comparisons/matrix_RE.txt
+++ b/comparisons/matrix_RE.txt
@@ -167,6 +167,7 @@ RunEGamma2018Bwf136p862 136.862_RunEGamma2018B+*/step3.root reRECO
 RunDoubleMuon2018Bwf136p863 136.863_RunDoubleMuon2018B+*/step3.root reRECO
 RunJetHT2018Bwf136p864 136.864_RunJetHT2018B+*/step3.root reRECO
 RunMET2018Bwf136p865 136.865_RunMET2018B+*/step3.root reRECO
+RunEGamma2018Cwf136p874 136.874_RunEGamma2018C+*/step3.root reRECO
 RunJetHT2018Cwf136p876 136.876_RunJetHT2018C+*/step3.root reRECO
 RunEGamma2018Dwf136p886 136.886_RunEGamma2018D+*/step3.root reRECO
 RunJetHT2018Dwf136p888 136.888_RunJetHT2018D+*/step3.root reRECO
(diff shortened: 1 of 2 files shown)


smuzaffar:
please test
@slava77 , now you should be able to test this too. Just doing `please test`  or `please test with cms-sw/cmssw#NUM` will allow to tests these changes during PR comparisons.

slava77:
> @slava77 , now you should be able to test this too.

This is nice.
May I suggest to modify the PR acknowledgment message (the bot's "A new Pull Request was created") to have this note about the possible commands, like "cms-bot commands are listed here" showing up in the cms-sw/cmssw PRs.
Thank you.

smuzaffar:
+externals
sure @slava77 , we will update the message

smuzaffar:
@slava77 , are these `Total failures: 131` expected?

fabiocos:
@smuzaffar this is a recent issue that has appeared apparently between CMSSW_10_6_X_2019-05-09-1100 and CMSSW_10_6_X_2019-05-09-2300, and is visible in any PR test since then. This is why I have tested reverting my relmon changes, anyway as I was expecting they are unrelated to this (i.e. reverting does not make the differences disappear). Funnily these differences are not seen in the new DQM bin-to-bin comparison

slava77:
some DT plots now have nan values in the histograms. These are showing up as failures in comparisons. Perhaps it's worth to review the logic of comparing values to let a nan be equal to another nan. This would get rid of the differences.

fabiocos:
@slava77 I am looking into it, NaNs are these since a while, I  am cross checking whether my update of relmon has just modified its behaviour wrt to them (it should have not but...)

davidlange6:
nans that get used in the comparison used to trigger differences in relmon (because nan!=nan)

fabiocos:
ok, it depends on my PR's that now picks all the expected bins, I have to add some protection for this

davidlange6:
Maybe we can fix the nans?




On 13 May 2019, at 14:44, Fabio Cossutti <notifications@github.com<mailto:notifications@github.com>> wrote:


ok, it depends on my PR's that now picks all the expected bins, I have to add some protection for this

fabiocos:
I would say that this should be done irrespectively of any protection...

slava77:
can this be merged (independent of the conclusion on the RelMon feature discussion)?

fabiocos:
+1

PR #1144, opened 2019-05-24
Escape an argument to support space separated values in DQM bin by bin tool

andrius-k:
This PR fixes an issue where multiple PRs separated by a space inside PULL_REQUESTS argument causes the tool to misbehave.


andrius-k:
please test

PR #1156, opened 2019-06-17
enable code-format for latest releases with new config_tag

smuzaffar:
@fabiocos , this should enable code-format for latest release cycles.

### DIFF
--- a/run-pr-code-checks
+++ b/run-pr-code-checks
@@ -7,10 +7,15 @@ export USER_CODE_CHECKS=$2
 BUILD_NUMBER=$3
 DRY_RUN=$4
 REPOSITORY=$5
-CODE_FORMAT=$6
-CODE_CHECKS=$7
-if [ "X$CODE_CHECKS" = "X" ] ; then CODE_CHECKS=true ; fi
-if [ "X$CODE_FORMAT" = "X" ] ; then CODE_FORMAT=false ; fi
+CODE_CHECKS=true
+CODE_FORMAT=false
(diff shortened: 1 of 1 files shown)


fabiocos:
@smuzaffar thanks, this will be anyway active only for PRs in the master branch, right? There is no code-checks enables for older cycles. Everything is now driven by $CMSSW_BASE/config/config_tag, this is the tag of the used snapshot of https://github.com/cms-sw/cmssw-config right?
What I do not understand is the master version of the SCRAM setup needs to be merged if the tag is up to date. For any tag posterior in sync with master this should be useless, am I correct?

smuzaffar:
@fabiocos , yes code-checks only runs by default for master branch but one can explicitly run it for non-master branch by issuing `code-checks` comment.
This change will allow to run `code-format` for old release cycles if config_tag is correct. For example when we have CMSSW_11_1_X or CMSSW_12_0_X then we still want to keep running (of course on demand only) code-format for CMSSW_11_0_X ... right?

fabiocos:
@smuzaffar ok, I get the point. If we want to keep the possibility to run the procedure on older releases I see the logic of the code. Of course doing it now will force clang-format on older code, so one should be even more careful in doing so.

smuzaffar:
+externals
This has been tested for https://github.com/cms-sw/cmssw/pull/27225

smuzaffar:
@fabiocos , feel free to merge. As soon it is merged code-format will be enabled for PR

fabiocos:
+1

PR #1170, opened 2019-07-19
Add test wf after renaming 2023DXX into 2026DXX

fabiocos:
The PR cms-sw/cmssw#27488 has renamed the Phase 2 scenarios 2023DXX into 2026DXX . This PR updates the bot comparison matrix so as to account for these new names and allow correct comparisons in 11_0_X branch after July 17.


fabiocos:
please test

fabiocos:
@slava77 could you please verify that this update is along the scheme that you maintain? Or suggest an updated version?

slava77:
> @slava77 could you please verify that this update is along the scheme that you maintain? Or suggest an updated version?

I have #1171

fabiocos:
@slava77 ok, as the last previous-to-last update was before the last pre-release no need to keep that state around, I think the difference between your PR and mine is just that

PR #1174, opened 2019-07-22
update list of watched packages - deguio

deguio:


deguio:
one of the packages I've listed doesn't exist yet, but it will appear after [PR 27582](https://github.com/cms-sw/cmssw/pull/27582)
is merged. Please let me know if you have different suggestions.

PR #1175, opened 2019-07-23
Update watchers.yaml

forthommel:
More PPS packages monitored!


forthommel:
Thanks for the prompt reaction, @smuzaffar !

PR #1179, opened 2019-07-31
reco comparisons: add a maping for D46, only for ttbar wf 22034

slava77:
the 22034 wf was added to the standard set of tests in cms-sw/cmssw#27555 starting from CMSSW_11_0_X_2019-07-31-1100 and now triggers reco comparisons warnings due to a missing map.
This PR updates the mapping file.


smuzaffar:
please test

smuzaffar:
+externals

PR #1183, opened 2019-08-06
reco comparisons: update maps for Run3 workflows

slava77:
Add some limited mappings for 2021, 2023 and 2024 workflows.

this should also clear the warnings in comparisons appearing since CMSSW_11_0_X_2019-08-04-0000


slava77:
please test

mrodozov:
can we try testing with the wf in question maybe ?

smuzaffar:
+externals

fabiocos:
@slava77 related to this update and your comment https://github.com/cms-sw/cmssw/pull/27678#discussion_r310370826 : is your proposal to add for instance wf 12434.0 ? If i am not wrong at present this is differing from 11634.0 essentially only for the conditions

slava77:
> @slava77 related to this update and your comment [cms-sw/cmssw#27678 (comment)](https://github.com/cms-sw/cmssw/pull/27678#discussion_r310370826) : is your proposal to add for instance wf 12434.0 ? If i am not wrong at present this is differing from 11634.0 essentially only for the conditions

Sure, just copy-pasting my comment there:
> given some failures in 2023,2024 in some other tests, should we add at least one to the limited set?

PR #1192, opened 2019-08-23
Addition to the list of watchers

apsallid:
Could you please add me to the list of watchers? Thanks.


mrodozov:
please test
just for the record, and then yes

PR #1197, opened 2019-08-29
run static checks with -DEDM_ML_DEBUG -w for PR tests

smuzaffar:


smuzaffar:
+externals

fabiocos:
@smuzaffar is this meant to probe code that otherwise would be ignored in regular runs?

smuzaffar:
yes but only for static checker. clang and gcc compilations are done without EDM_ML_DEBUG.

PR #1203, opened 2019-09-25
Ib rotate weeks improved

tulamor:
Variables initialized automatically using config.map instead of hardcoded values


smuzaffar:
looks good

PR #1210, opened 2019-10-15
Install gfal2-python as part of rucio installation

clelange:
The scripts developed by @dciangot require the [gfal2-python](https://pypi.org/project/gfal2-python/) package to work so they should be installed when installing rucio.

This is related to https://github.com/cms-sw/cmsdist/issues/5199 (and could resolve it).

FYI @belforte @ericvaandering


clelange:
Closing as per discussion in https://github.com/cms-sw/cmsdist/issues/5199

PR #1216, opened 2019-10-25
Update of PPS/TOTEM data formats accessors

forthommel:
Dropped all `get` prefixes from PPS/TOTEM leaves, following https://github.com/cms-sw/cmssw/pull/28252


comparisons/validate.C
```diff
@@ -984,59 +984,59 @@ void validateEvents(TString step, TString file, TString refFile, TString r="RECO
       //CTPPS
       tbr="TotemFEDInfos_totemRPRawToDigi_RP_";
       plotvar(tbr+recoS+".obj@.size()");
-      plotvar(tbr+recoS+".obj.getFEDId()");
-      plotvar(tbr+recoS+".obj.getOptoRxId()");
-      plotvar(tbr+recoS+".obj.getFSize()");
+      plotvar(tbr+recoS+".obj.fedId()");
```
slava77: @forthommel  thanks for looking at this as well. Did you close the PR intentionally?

This code will have to support running diffs on the old releases as well as the new ones.
The approach to satisfy this is to *add* the new methods.

PR #1220, opened 2019-11-01
update reco comparisons: new products, automate handling of range scans

slava77:
- TICL multiclusters (development)
- add pat::Muon inverseBeta
- automate size detection for object vector/array scans; use instead of hard-coded ranges


slava77:
@cmsbuild please test

slava77:
the logs in validateJR look healthy (no crashes and the range scans are now consistent with the actual sizes of stored data instead of previous guesses).
I think that this update is good to go.

smuzaffar:
+external

PR #1223, opened 2019-11-18
Remove geoff-smith from watchers.yaml

makortel:
As requested in https://github.com/cms-sw/cmssw/pull/28417#issuecomment-555106826

FYI @geoff-smith


geoff-smith:
Thanks!

PR #1228, opened 2019-11-26
update mapping file for reco comparisons

slava77:
update mapping for reco comparisons:
- for local tests
    - ppremix workflow for 2021
    - add run3 test wf 11634.7
- for the standard jenkins tests:
    - add/fix a map for fastsim worfklows using PAT/RECO process names <== THIS IS A BUGFIX, which left the standard jenkins tests reco comparisons missing


slava77:
@cmsbuild please test

fabiocos:
@slava77 I believe a further update is needed to full exploit cms-sw/cmssw#28487, by adding wf 23234.0, am I correct?

smuzaffar:
+externals

fabiocos:
@smuzaffar probably we can integrate this and add another one with the further update

slava77:
> @slava77 I believe a further update is needed to full exploit [cms-sw/cmssw#28487](https://github.com/cms-sw/cmssw/pull/28487), by adding wf 23234.0, am I correct?

right, I will do it separately.

Thanks for merging this already.

PR #1241, opened 2019-12-19
reco comparisons updates for HGCAL

slava77:
- HFNose hits and layer clusters
- hgcalMultiClusters: pick a correct collection ncollection name, updated in 102X already


slava77:
@cmsbuild please test

PR #1250, opened 2020-01-22
Status functions to use for pull requests

mrodozov:
Given github pull request object exists, have functions to add new status and get overall status from the "combined" status. For tests see prev commits. 
Keep this one as a reference and add what more we need


smuzaffar:
-1

PR #1256, opened 2020-01-29
Externals stats on Elastic

mrodozov:
This adds changes for
1. Add summary function get_summary_stats_from_dictionary
Records are as expected
https://es-cmssdt.cern.ch/kibana/app/kibana#/discover?_g=()&_a=(columns:!(_source),index:AW_svgachldp7JMpTV51,interval:auto,query:(match_all:()),sort:!('@timestamp',desc))
2. Add script to be used by Jenkins
The result of es_send_resource_stats was validated against result produced with the old version,
jsons (values of sdata, see line 248) are the same

### DIFF
--- a/es_externals_stats.py
+++ b/es_externals_stats.py
@@ -1,0 +1,83 @@
+#!/usr/bin/env python
+from __future__ import print_function
+from cmsutils import percentile
+from es_utils import send_payload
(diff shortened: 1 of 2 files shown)


mrodozov:
meh too many commits

PR #1258, opened 2020-01-29
reco comparisons:  monitor hcal reduced hits in miniAOD; map entry for 136.793_RunDoubleEG2017C

slava77:
- monitor hcal reduced hits in miniAOD
- add a map entry for 136.793_RunDoubleEG2017C, which now appears in the short matrix tests running by jenkins


slava77:
@cmsbuild please test

smuzaffar:
+externals

PR #1259, opened 2020-02-03
Add  RecoMTD/TimingIDTools under upgrade and reco

mrodozov:
Related to this https://github.com/cms-sw/cmssw/pull/28815


mrodozov:
@slava77 this should go in upgrade (following the other RecoMTD/* pkgs location) ?

slava77:
> @slava77 should this go in upgrade (following the other RecoMTD/* pkgs location) ?

I think that it should be in reco and upgrade.

mrodozov:
please test

PR #1262, opened 2020-02-10
Rename comment-gh-pr comment-github-pullrequest. Add comment_gh_pr.py.

gartung:
This allows custom comment_gh_pr module/function to be added to repos directory.


comment_gh_pr.py
```diff
@@ -0,0 +1,8 @@
+from github import Github
+
+def comment_gh_pr(repo, pr, msg):
+      from os import environ
+      gh = Github(login_or_token=environ['GITHUBTOKEN'], retry=3)
```
smuzaffar: @gartung , is not it going to break cms bot test? here it assumes that GITHUBTOKEN env is set which in original comment-gh-pr we were using `repo_config.GH_TOKEN`

gartung: Sorry I forgot I adapted to using an environment variable set from a stored Jenkins secret. I will change it to the CMS convention.

gartung:
A graduate student on Mu2e added his own process_pr.py to their repo definition and he also wanted to implement his own comment_gh_pr function. He is using the github commit status functions and wanted to make his changes orthogonal to LArSoft changes. I never thought of doing that because LArSoft has 20 repos.

comment_gh_pr.py
```diff
@@ -0,0 +1,8 @@
+from github import Github
+
+def comment_gh_pr(repo, pr, msg):
+      from os import environ
+      gh = Github(gh = Github(login_or_token=open(expanduser(repo_config.GH_TOKEN)).read().strip()))
```
smuzaffar: I do not like this, here it is making use of expanduser and repo_config without any ref to these. 
I would suggest keep comment-gh-pr as it is and just add extra function in github_utils.py e.g.

```
def comment_gh_pr(repo, pr, msg):
      gh = Github(login_or_token=get_gh_token(repo))
      repo = gh.get_repo(repo)
      pr   = repo.get_issue(pr)
      pr.create_comment(msg)  
```

gartung: OK. I just wanted to share his idea of using your own comment_gh_pr so you can do additional things like set commit statuses.

smuzaffar:
-1
please add this comment_gh_pr function in github_utils.py , no need to have 7 files updated for this simple change

PR #1277, opened 2020-03-24
Update of categories.py as asked by PPD

silviodonato:
Update of categories.py as asked by PPD.
Could you double check @srimanob ?
cc: @franzoni, @zhenhu, @wajidalikhan


srimanob:
@silviodonato 
Good for me, thanks.

PR #1278, opened 2020-03-25
Clean up old release managers

fwyzard:
Remove previous release managers from the cms-bot repository.


smuzaffar:
thanks for cleanup @fwyzard, I have no objection. We kept them there in case current release manager is not around and we have to build urgent release. 
@silviodonato, @davidlange6, @fabiocos and @slava77 do you have any objection?

slava77:
I was added for a period of time in 2018 when Fabio was not available and needed/used this role just in that period and we had other volunteers to lead in some other interim times.

So, it's fine for me.
 I will have less occasional CMS developers at large assume that I'm responsible for the builds ;)

fabiocos:
@slava77 I believe you were added quite before my mandate, as seen in the history, and we never updated this, but it does not matter.

As far as I am concerned no problem in being removed, if  it is ok for @silviodonato and @smuzaffar,  fewer e-mails to check . I just suggest to ensure that Silvio has some backup available to act as a release manager in case of absence (in principle @kpedro88 should have the rights). In case of need for limited periods keep in mind I am also available to help.

silviodonato:
I agree with @fabiocos , I would like to have at least one backup person in case of problems/emergency . If @kpedro88 has already the rights, I'm ok with the change.

fabiocos:
it looks like @kpedro88 was removed from the list, so not sure there is anybody else left beyond @silviodonato after this PR, @smuzaffar may confirm.

fwyzard:
I guess I should add Fabio back :-)

categories.py
```diff
@@ -8,7 +8,7 @@
 
 authors = {}
 GITHUB_BLACKLIST_AUTHORS = []
-CMSSW_L1 = ["davidlange6", "fabiocos", "silviodonato"]
+CMSSW_L1 = ["silviodonato"]
```
fwyzard:
```suggestion
CMSSW_L1 = ["fabiocos", "silviodonato"]
```

slava77:
> @slava77 I believe you were added quite before my mandate, as seen in the history, and we never updated this, but it does not matter.

Ah, I got the year wrong, it was 2015 HI run when I used any release building role

PR #1280, opened 2020-03-26
updates in reco Comparisons: monitor PUPPI jets

slava77:
- prod-like datasets for phase-1/2 workflows (used for private tests so far)
- monitor puppi jets in RECO to support changes added in cms-sw/cmssw#28936 
- add a set of jet tag plots for AK8 with pt  > 200 GeV, which roughly corresponds to the range where these are used for physics (while many jets are left untagged for lower momentum range)


slava77:
@cmsbuild please test

slava77:
It looks like the reco comparisons finished OK; this validates the update in validate.C

smuzaffar:
+externals

PR #1282, opened 2020-03-27
Add fabiocos to the watchers list

fabiocos:
Define a list of packages I want to keep monitoring, regardless of my presence in other roles.


fabiocos:
@smuzaffar I believe this addition is not conflicting with other roles within cms-bot

PR #1283, opened 2020-03-31
updates in reco comparisons

slava77:
- matrix mapping file update for patatrack workflows (still just for private tests)
- updating reco validation to match recent developments
    - tau IDs from cms-sw/cmssw#28417
    - met unclustered variable


slava77:
@cmsbuild please test

smuzaffar:
@slava77 , are these comparison differences expected?

slava77:
> @slava77 , are these comparison differences expected?

all differences are not related to this PR; they appear in LogWarnings differences.

I checked one case in wf 12434.0 step3
```
%MSG-e ExcessiveTime:  SiPixelClusterProducer:siPixelClustersPreSplitting  31-Mar-2020 04:17:48 CEST Run: 1 Event: 1
ExcessiveTime: Module used 764.556 seconds of time which exceeds the error threshold configured in the Timing Service of 600 seconds.
```
is present in the baseline but not in the PR test.
I've seen something like this before from some transient effects in the testing infrastructure.

Based on the tests, I think that this PR is good to go.

smuzaffar:
+externals

PR #1292, opened 2020-05-01
reco comparisons: TICL monitoring: tracksters with multiclusters, TICL and PF Candidates

slava77:
updates in validate.C used for reco comparisons. The main part is the addition of monitoring for TICL reconstruction, which is available in the default workflows since 11_1_0_pre6.

### DIFF
--- a/comparisons/validate.C
+++ b/comparisons/validate.C
@@ -1132,7 +1132,44 @@ void hgcalMultiClusters(TString cName ){
   plotvar(bObj+".eta()");
   plotvar(bObj+".phi()");
   plotvar("log10("+bObj+".energy())");
-  plotvar(bObj+".time()");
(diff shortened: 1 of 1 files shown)


slava77:
@cmsbuild please test

slava77:
it may be better to merge this only after cms-sw/cmssw#29616 is merged to avoid random differences in comparisons

slava77:
@cmsbuild please test

after CMSSW_11_1_X_2020-05-04-2300 the differences in reco should go away

smuzaffar:
@slava77 , is ready to go in now?

slava77:
> @slava77 , is ready to go in now?

yes, it's ready.
Thanks for checking.

smuzaffar:
+externals

PR #1298, opened 2020-05-06
remove GurpreetSinghChahal for now. He need to first accept the membership

smuzaffar:


smuzaffar:
@qliphy , I am removing @GurpreetSinghChahal  fow now. He need to accept the membership request first (which is open now for few days). Once he become a member then you can add him in to the team

PR #1302, opened 2020-05-13
reco comparisons: add map entries for Run2 UL reminiAOD workflows

slava77:


slava77:
@cmsbuild please test with cms-sw/cmssw#29826

the extra PR is needed for the tests to complete successfully, else the IB itself is broken

slava77:
@smuzaffar 
I think that this can go in by itself; but the validation is also confirmed based on the joint test with  cms-sw/cmssw#29826

smuzaffar:
+externals

PR #1309, opened 2020-05-19
Filter 'edm::ConcurrentHadronizerFilter<Pythia8Hadronizer' in modules2statics report

gartung:
The only way to avoid reporting the link to statics in ```edm::ConcurrentHadronizerFilter<Pythia8Hadronizer, gen::ConcurrentExternalDecayDriver>::filter()```  is to mark all statics linked through the base class as safe. To avoid doing this just filter edm::ConcurrentHadronizerFilter<Pythia8Hadronize  from the modules2statics report.

### DIFF
--- a/run_clang_static_analysis
+++ b/run_clang_static_analysis
@@ -27,7 +27,7 @@ fi
 ./create_statics_esd_reports.sh
 ls -ltrh
 tar cvfz dumper-checker.tgz *.txt*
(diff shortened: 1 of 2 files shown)


gartung:
please test

makortel:
I wonder if we should start to consider a separate file listing the cases to be filtered out instead of a pipe of greps.

gartung:
The original file is still kept.

gartung:
> I wonder if we should start to consider a separate file listing the cases to be filtered out instead of a pipe of greps.

Because the output is piped to negative greps.

smuzaffar:
@gartung , `egrep -v -f statics-filter1.txt` should work where `statics-filter1.txt` could have
```
edm::RootInputFileSequence::initTheFile()
PrintGeomInfoAction::update()
edm::ConcurrentHadronizerFilter<Pythia8Hadronizer 
```

makortel:
Thanks!

PR #1319, opened 2020-06-02
reco comparisons: add plots for slimmedMETsNoHF

slava77:
- in miniAOD plots: add monitoring for slimmedMETsNoHF
- in the matrix map add maps for D49 workflows with PU, currently used only in private tests


slava77:
@smuzaffar 
the update here works as expected.
On the other hand, I think I will have another plot to add later today (unrelated to the one added here), which will go to this PR if it's not merged yet.

smuzaffar:
+externals
@slava77 , I am merging it now

PR #1322, opened 2020-06-03
reco comparisons: bugfix for the map for OldDigi_2026D49 and add FTLRecHit position plot

slava77:
- recently added mapping for wf 23234.1001 had a wrong file name; fixed here
- cms-sw/cmssw#29803 has a new variable useful for monitoring; added here


slava77:
@cmsbuild please test

to confirm that 23234.1001 reco diff will appear in the outputs

smuzaffar:
+externals

silviodonato:
> @cmsbuild please test
> 
> to confirm that 23234.1001 reco diff will appear in the outputs

@slava77 the comparison test shows `Reco comparison results: 0 differences found in the comparisons`. Is it ok?

slava77:
> @slava77 the comparison test shows `Reco comparison results: 0 differences found in the comparisons`. Is it ok?

this summary is for actual differences in distributions.
This PR does not modify anything produced in reco (CMSSW),
in most cases this can not introduce differences.

I was not precise in my wording, I meant that the directory for diffs will appear as it did in 
https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_2_X_2020-06-02-2300+98c400/36798/validateJR/all_OldVSNew_TTbar14TeV2026D49OldDigiwf23234p1001/

silviodonato:
+1

PR #1324, opened 2020-06-09
reco comparisons: update in caloTowers

slava77:
- move caloTowers plotting to a separate method 
- add a plot for PFTowers (present in heavyIons workflows only); triggered by changes in cms-sw/cmssw#29948 
- bugifx mass2 method name, should be massSqr (since ever)

no changes are expected in the reco comparisons summaries

### DIFF
--- a/comparisons/validate.C
+++ b/comparisons/validate.C
@@ -353,6 +353,36 @@ void jets(TString type,TString algo){
   }
 }
 
+void caloTowers(TString br, bool addHEP17 = false){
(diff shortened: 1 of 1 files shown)


slava77:
@cmsbuild please test

smuzaffar:
+externals
@slava77 , is the ready to go in?

slava77:
>  @slava77 , is the ready to go in?

yes it is, 
the test with a CMSSW PR that benefits from this one showed the differences for newly added plots

PR #1328, opened 2020-06-23
make use of new log-deps option

smuzaffar:
to test the new log-dep option of cmsBuild ( https://github.com/cms-sw/pkgtools/pull/160 )


smuzaffar:
+externals

PR #1339, opened 2020-07-07
Add gouskos and mariadalfonso as xpog

silviodonato:
Adding the Loukas Gouskos and Mariarosaria D'Alfonso as new xpog manager.


categories.py
```diff
@@ -42,6 +42,7 @@
   "franzoni":         ["operations"],
   "fwyzard":          ["heterogeneous", "hlt"],
   "ggovi":            ["db"],
+  "gouskos":          ["xpog"]
```
smuzaffar, line 45: missing `,` at the end

PR #1340, opened 2020-07-07
reco comparisons: plots for JPT jets;  packed candidates from heavy ion miniAOD

slava77:
- add plots for JPT jets AK4 (forgotten post-migration from AK5 since a while). Needed for cms-sw/cmssw#30430
- add plots of counts of packedCand refs from packedCandidateMuonID (heavy ions re-miniAOD). Needed for/after cms-sw/cmssw#30253 
- hiPixelTracks as packed candidates (heavy ions re-miniAOD). Needed for/after cms-sw/cmssw#30313

### DIFF
--- a/comparisons/validate.C
+++ b/comparisons/validate.C
@@ -2841,11 +2841,20 @@ void validateEvents(TString step, TString file, TString refFile, TString r="RECO
       }
     }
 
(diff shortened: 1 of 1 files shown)


slava77:
@cmsbuild please test

smuzaffar:
+externals

PR #1341, opened 2020-07-08
Check tag pattern before using it

mrodozov:


create-new-data-pr.py
```diff
@@ -69,7 +69,7 @@ def update_tag_version(current_version=None):
   new_tag = last_release_tag # in case the tag doesnt change
   if create_new_tag:
       nums_only = last_release_tag.strip('V').split('-')
-      first, sec, thrd = tuple(nums_only)
+      first, sec, thrd = tuple(nums_only)[:3]
```
smuzaffar, line 72: @mrodozov , what if https://github.com/cms-sw/cms-bot/blob/39982b96b37362d9ee523dd46e8ae7ae07928a07/create-new-data-pr.py#L48 doesnot return a tag with VNN-NN-NN format? what if someone creatse a tag with wrong format? I think you also need to fix the logicin https://github.com/cms-sw/cms-bot/blob/39982b96b37362d9ee523dd46e8ae7ae07928a07/create-new-data-pr.py#L48 so that last_release_tag is always of form Vnn-nn-nn

PR #1356, opened 2020-08-04
Add --ibeos to runTheMatrix.py options

gartung:


run-ib-igprof
```diff
@@ -8,7 +8,7 @@ if [ "X$EVENTS" = "X" ] ; then EVENTS=1000; fi
 for prof in ${PROFILES} ; do
   mkdir -p $WORKSPACE/igprof/${prof}
   cd $WORKSPACE/igprof/${prof}
-  runTheMatrix.py $WORKFLOWS --command "-n $EVENTS --profile $prof" 2>&1 | tee runTheMatrix.log
+  runTheMatrix.py $WORKFLOWS --command "--ibeos -n $EVENTS --profile $prof" 2>&1 | tee runTheMatrix.log
```
smuzaffar, line 11: although no harm but there is not need to use `--ibeos` here. Jenkins jobs sets the correct CMS_PATH.
By the way, `--ibeos` option should be outside e.g. `runTheMatrix.py $WORKFLOWS --ibeos --command "-n $EVENTS ....`

PR #1359, opened 2020-08-05
Output Singularity environment variables and command to run_singularity.sh  if enviromnent variable OUTPUTCMD=1 is set.

gartung:
Running the singularity command from its own script seems to avoid the segfault when running igprof under it.


gartung:
Didn't work.

PR #1363, opened 2020-08-17
If igprof is being run do no set ulimit.

gartung:
Testing showed that commenting out the ulimit commands allowed igprof to run under Singularity. 

Used the Jenkins job parameter/ environment variable ADDITIONAL_TEST_NAME to skip running ulimit when ADDITIONAL_TEST_NAME=igprof.


gartung:
The problem seems to be with setting any hard ulimits. It looks like the jenkins client launcher script also finds the hard limits on the node and sets them in the Jenkins client environment.

PR #1364, opened 2020-08-17
When starting jenkins client on cmsdev machines don't run ulimit -H.

gartung:
Running ulimit -H causes igprof to segfault when run under Singularity.

### DIFF
--- a/docker_launcher.sh
+++ b/docker_launcher.sh
@@ -1,6 +1,7 @@
 #!/bin/bash -ex
 if [ "X$ADDITIONAL_TEST_NAME" = "Xigprof" ] ; then
   if ulimit -a ; then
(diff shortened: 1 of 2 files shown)


smuzaffar:
@gartung , did you test this on cmsdev? Can you run limit command by hand and run igprof under singularity and check if it actually fails?

gartung:
I ran 
```
$WORKSPACE/cms-bot/docker_launcher.sh $WORKSPACE/run.sh 
```
on cmsdev34 directly and it ran without segfault. The only difference I could see was that Jenkins client was launched with the ulimit pre commands.

gartung:
Superceeded by
https://github.com/cms-sw/cms-bot/pull/1365

PR #1376, opened 2020-09-10
Generate the RECO event loop summary using the igprof text reports

gartung:


smuzaffar:
@gartung , are you happy with this change, I mean have you already testing in the jenkins job?

PR #1377, opened 2020-09-10
Use cmssw-pr-test-config to select profiling workflow

gartung:


gartung:
@smuzaffar  This is setting WORKFLOWS=-l PROFILING_WORKFLOWS. 
https://cmssdt.cern.ch/jenkins/job/ib-run-igprof/11345/parameters/
I need to resubmit to fix this.

PR #1379, opened 2020-09-11
bugfix in process name used for reminiAOD for HI MC reco comparisons mapping

slava77:
apparently it took more than 2 months to discover that the reco comparisons plotter is looking at a wrong process name.


slava77:
@cmsbuild please test with cms-sw/cmssw#30898

so that there are some differences

slava77:
In the comparison diff summary https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_11_2_X_2020-09-11-1100+11aab6/38779/validateJR.html
I now see the workflow 158.01 show up as expected

WF # | Differences
-- | --
all_OldVSNew_HydjetQB12ppRECOin2018reMINIAODwf158p01 | 205

@mrodozov @smuzaffar 
I think that this can be merged

PR #1381, opened 2020-09-14
Add  PhysicsTools/IsolationAlgos to reconstruction

silviodonato:
From https://github.com/cms-sw/cmssw/pull/31455


silviodonato:
@cms-sw/reconstruction-l2 do you agree?

slava77:
> @cms-sw/reconstruction-l2 do you agree?

I can't recall why (or if) I resisted  to add it in the past.
All dependent packages are in the reco category and analysis category is/was typically left for code used in later stages.

PR #1390, opened 2020-10-01
Create TkDQM group and add it to watchers

arossi83:
As the title already says this PR is meant to create a group for Tracker DQM conveners and add this group to the watchers of the relevant packages.
@sroychow is aware of this!


smuzaffar:
any objectiuons @sroychow ?

sroychow:
@smuzaffar No objections. Thank you.

PR #1395, opened 2020-10-10
Add more CUDADataFormats packages for the Patatrack integration

fwyzard:
See
  - cms-sw/cmssw#31703: Patatrack integration - common data formats (5/N)
  - cms-sw/cmssw#31704: Patatrack integration - calorimeters shared code (6/N)
  - cms-sw/cmssw#31719: Patatrack integration - ECAL local reconstruction (7/N)
  - cms-sw/cmssw#31720: Patatrack integration - HCAL local reconstruction (8/N)
  - cms-sw/cmssw#31721: Patatrack integration - Pixel local reconstruction (9/N)
  - cms-sw/cmssw#31722: Patatrack integration - Pixel track reconstruction (10/N)
  - cms-sw/cmssw#31723: Patatrack integration - Pixel vertex reconstruction (11/N)


categories_map.py
```diff
@@ -849,11 +849,19 @@
     "Validation/Shashlik",
   ],
   "heterogeneous": [
-    "CUDADataFormats/Common",
     "CUDADataFormats/BeamSpot",
+    "CUDADataFormats/CaloCommon",
```
slava77, line 853: if these are supposed to be used in reco CUDA, I think that a reco signature is needed as well.

smuzaffar: @fwyzard , any idea if these will be used by reco CUDA?

fwyzard: yes

smuzaffar: can you please update the PR to add these under reconstruction too

fwyzard: shall I also add under "heterogeneous" all packages that do or will contain any kind of CUDA code ?

smuzaffar: that is a good question, dependency of a category X should not be responsibility of that category X.

fwyzard: Sorry, what I meant is: should heterogeneous (currently, only CUDA) code be signed by the `heterogeneous` category ?
If the answer is "no", then what about the data formats just added under `reconstucrtion` ?

makortel: I have a vague recollection that the last time this was discussed (sorry no reference) the outcome was **not** to add `heterogeneous` on every package with CUDA code, but to be assigned by other L2s when necessary. I'm fine with moving all other `CUDADataFormats` than `CUDADataFormats/Common` and `CUDADataFormats/StdDictionaries` only under `reconstruction`.

makortel: Let me add (post-merge) that after some further thought I think I contradicted myself in above (assign `heterogneous` on PRs on demand, but leave `CUDADataFormats` packages out from `heterogeneous`), and leaving all `CUDADataFormats` packages (also) under `heterogeneous` is more consistent.

PR #1404, opened 2020-10-21
Add CalibPPS/AlignmentRelative to alca

mrodozov:
see https://github.com/cms-sw/cmssw/pull/31629#issuecomment-710163243
@silviodonato , @jan-kaspar only in alca, right ?


jan-kaspar:
> see [cms-sw/cmssw#31629 (comment)](https://github.com/cms-sw/cmssw/pull/31629#issuecomment-710163243)
> @silviodonato , @jan-kaspar only in alca, right ?

This would be my guess, but I am not an expert.

silviodonato:
+1
@mrodozov yes, right

PR #1418, opened 2020-11-14
prep to build geant4 IB on 15.11 at 1100

mrodozov:
after the next Geant4 IB is built (today at 23) another one with different vecgeom setup needs to be build.
This is a prep for triggering geant4 IB sunday 15.11 at 1100


smuzaffar:
@mrodozov , days are from 0 to 6, so day 7 is not going to work.
Also there is no jenkins job to run at 11h00 on Sunday. If you want to build a special IB then just trigger by hand

mrodozov:
ah the full IB flag will do :)

PR #1422, opened 2020-11-19
reco comparisons updates for: TICL plots; DD4HEP short matrix wfs

slava77:
- TICL-related monitoring was updated after cms-sw/cmssw#31906 was merged
- matrix map was extended to cover two new DD4HEP workflows added in cms-sw/cmssw#32096 (11624.911 and 11642.911)


slava77:
@cmsbuild please test

slava77:
From the test results, I think that the PR is working as expected:
- the warnings about 116{24,42}.911 workflow maps being missing are gone
- the validateJR logs show that the new TICL variables are plotted (defaults had no differences)

smuzaffar:
+externals

PR #1424, opened 2020-11-20
Stop running monitor on arm

mrodozov:
this runs on arm and intel
only concern is the missing wf_stats files not created by the monitor script
from 30 failing tests only 3 failed when ran without monitor indeed

### DIFF
--- a/RelValArgs.py
+++ b/RelValArgs.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 from __future__ import print_function
 import re
-from os import environ
+from os import environ, uname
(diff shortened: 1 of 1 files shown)


smuzaffar:
@mrodozov , did you check if this will work for python3?

mrodozov:
it works
```
Python 3.8.2 (default, May  7 2020, 20:05:56) 
[GCC 8.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from os import uname
>>> last = uname()[-1]
>>> print(last)
aarch64
```

PR #1433, opened 2020-12-01
reco comparisons driver: check that both input root files are available before running validate.C

slava77:
while testing https://github.com/cms-sw/cmssw/pull/32325#issuecomment-735484569
`Reco comparison had 1 failed jobs` showed up somewhat unexpectedly.
It was triggered by a request to include a workflow in the test which does not have a reference. These unpaired inputs should not be used to run validate.C (fwlite-based reco comparisons).

this PR should resolve the issue


slava77:
@cmsbuild please test workflow 140.5611 with cms-sw/cmssw#32325

smuzaffar:
+externals

PR #1436, opened 2020-12-01
treat QALOG as a text file in run-pr-comparisons

slava77:
an occasional special character in logRootQA file can break parsing of the summary of the PR tests.
One example is in https://github.com/cms-sw/cmssw/pull/32330#issuecomment-736813502

Screen Shot 2020-12-01 at 2 42 39 PM

A simple test with the same file downloaded locally:
```
>grep SUMMARY logRootQA.log
Binary file logRootQA.log matches

>grep -a SUMMARY logRootQA.log
SUMMARY You potentially added 2947 lines to the logs
SUMMARY Reco comparison results: 0 differences found in the comparisons
SUMMARY DQMHistoTests: Total files compared: 28
...
```


slava77:
@cmsbuild please test with cms-sw/cmssw#32330

PR #1440, opened 2020-12-04
Add functions to get externals stats

mrodozov:
this gets the externals stats from elastic search for the last 30 days by default from externals_stats_summary_testindex*

### DIFF
--- a/es_utils.py
+++ b/es_utils.py
@@ -267,3 +267,26 @@ def es_send_external_stats(stats_dict_file_path, opts_dict_file_path, cpu_normal
   sdata["@timestamp"]=file_stamp*1000
   try:send_payload(es_index_name+"-"+week, es_doc_name, index_sha, json.dumps(sdata))
   except Exception as e: print(e.message)
+
+def getExternalsESstats(cmsdist='*', arch='*', lastNdays=30, page_size=0):
+    stats = es_query(index='externals_stats_summary_testindex*',
+                     query=format('cmsdist:%(cmsdist_branch)s AND architecture:%(architecture)s',
+                                  cmsdist_branch=str(cmsdist),
+                                  architecture=arch),
+                     start_time=1000*int(time()-(86400*lastNdays)),
+                     end_time=1000*int(time()),scroll=True)
+    return stats['hits']['hits']
+
+# get a dict of stats with externals name as keys
+# create a default github file with stats so if elastic search fails,
+def orderStatsByName(externalsStats=None):    
+    namedStats = {}
+    for element in externalsStats:
+        ext_name = element["_source"]["name"]
+        if ext_name not in namedStats:
+            namedStats[ext_name] = list()
+        namedStats[ext_name].append(element["_source"])
+    # order by timestamp
+    for ext in namedStats:
+        namedStats[ext] = sorted(namedStats[ext], key=lambda x: x["@timestamp"], reverse=True)
+    return namedStats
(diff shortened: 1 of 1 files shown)


es_utils.py
```diff
@@ -267,3 +267,26 @@ def es_send_external_stats(stats_dict_file_path, opts_dict_file_path, cpu_normal
   sdata["@timestamp"]=file_stamp*1000
   try:send_payload(es_index_name+"-"+week, es_doc_name, index_sha, json.dumps(sdata))
   except Exception as e: print(e.message)
+
+def getExternalsESstats(cmsdist='*', arch='*', lastNdays=30, page_size=0):
+    stats = es_query(index='externals_stats_summary_testindex*',
```
smuzaffar, line 272: @mrodozov , we should now start using production index i.e no need to have _testindex in the name now

mrodozov: I'm changing the name in the uploading script, than a re-run should create the renamed indexes and after that erasing the old ones

smuzaffar: sounds good.

es_utils.py
```diff
@@ -267,3 +267,26 @@ def es_send_external_stats(stats_dict_file_path, opts_dict_file_path, cpu_normal
   sdata["@timestamp"]=file_stamp*1000
   try:send_payload(es_index_name+"-"+week, es_doc_name, index_sha, json.dumps(sdata))
   except Exception as e: print(e.message)
+
+def getExternalsESstats(cmsdist='*', arch='*', lastNdays=30, page_size=0):
```
smuzaffar, line 271: @mrodozov , may be its better to not use cmsdist. There will be no data for new release cycles. How about we pass (optional) a dict of package name and realversions e.g `{"root": "6.22.06", "geant4": "10.6", ....}`. I think before starting the build we can get the list of packages which will be built and then only get the stats of those packages

mrodozov: arch only works too and sorting can be done later over names (or even cmsdist branch if matching since it's in the result)

es_utils.py
```diff
@@ -267,3 +267,25 @@ def es_send_external_stats(stats_dict_file_path, opts_dict_file_path, cpu_normal
   sdata["@timestamp"]=file_stamp*1000
   try:send_payload(es_index_name+"-"+week, es_doc_name, index_sha, json.dumps(sdata))
   except Exception as e: print(e.message)
+
+def getExternalsESstats(arch='*', lastNdays=30, page_size=0):
+    stats = es_query(index='externals_stats_summary-*',
+                     query=format('architecture:%(architecture)s',
+                                  architecture=arch),
+                     start_time=1000*int(time()-(86400*lastNdays)),
+                     end_time=1000*int(time()),scroll=True)
```
smuzaffar, line 276: time() is not defined 
```
NameError: global name 'time' is not defined
```

mrodozov: I added it, it wasn't available in es_utils.py (and I was using another script to run that function which has it, so didn't noticed)

smuzaffar:
by the way, have you tested these functions?
This reults values from different runs, where are you calculating the avg. to be used by cmsBuild?

mrodozov:
es_utils was reading the most data available and sort it by time, but that was given the arch was used. then the pkgtools class was getting the first record which would the the latest and not the average. 
now as we dont use arch the latest could be any arch and I think the pkgtools class have to do it using the build params i.e. use the name and version and do the avg or use the latest available for that arch

smuzaffar:
I think we are accessing too much unnecessary data from ES. Try to improve the query and only get the records which make sense. If you decided to use only latest record the query for only one record. Also limit the packages names to the one which we have to re-build. For example if root package is changed then only hand full of packages will be re-build , so there is no need to ask for over 500 packages data.

mrodozov:
we can use a csv list which defaults to * if no argument is provided (and will get all available.) 
getting the exact names - also checked.

PR #1442, opened 2020-12-07
ROOT 622 IB in 2300

mrodozov:
I see the root622 is set to build in 24, but it doesn;'t as well as patatrack releases
should it be 23, or 00 ?


smuzaffar:
we do not need root 6.22 IBs any more. Native 11.3.X Ibs are enough.

PR #1466, opened 2021-01-13
Use tag names on commits from base branch

mrodozov:
this is listing the commits on the base branch (either master or main)
and when it tries to create new tag/release skips tag names which are tagging commits not from the base/default branch.
this is tested and its working, see 
https://github.com/npc-data/PhysicsTools-NanoAOD/tags
and 
https://github.com/npcbot/cmsdist/pulls
the tag https://github.com/npc-data/PhysicsTools-NanoAOD/releases/tag/V01-06-01 is tagging non base branch and it is skipped,
so the previous tag is taken 01-06-00 to create 01-07-00
For PRs against non base branches the job fails (to be dealt manually)

### DIFF
--- a/create-new-data-pr.py
+++ b/create-new-data-pr.py
@@ -40,7 +40,7 @@ if __name__ == "__main__":
   if data_repo_base_branch == "main":
       if "master" not in [branch.name for branch in data_repo.get_branches()]:
           data_repo.create_git_ref(ref='refs/heads/master', sha=data_repo.get_branch(data_repo_base_branch).commit.sha)
-
+  # non base branch releases to be dealt with manually, the job will fail
   if (data_repo_base_branch != "master") and (data_repo_base_branch != "main"):
       print("I do not know how to tag a non-master (non-main) branch %s" % data_repo_pr.base.ref)
       exit(1)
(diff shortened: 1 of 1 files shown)


smuzaffar:
@mrodozov , I think it would be simpler if you use https://docs.github.com/en/free-pro-team@latest/rest/reference/git#list-matching-references and use the last entry to get the latest tag (looks like tags are in order there e.g. https://api.github.com/repos/cms-data/RecoEgamma-ElectronIdentification/git/matching-refs/tags )

smuzaffar:
or https://api.github.com/repos/cms-data/RecoEgamma-ElectronIdentification/tags and use the first match `V*` element

mrodozov:
I can get the last tag with the pygithub api sorting the tag_name alphabetically(or numerically),
but ordering by tag name doesn't guarantees the latest tag (by name/number) will be from the default branch.
It does in the last case, yes, but not in general.
Unless we adopt a policy that only we know of type "let never have a special tag with the biggest possible tag number" 😅
See the other PR, which is doing that with only two lines

smuzaffar:
can we close this in favor of #1469

PR #1468, opened 2021-01-14
Reduction of the number of categories under 'analysis'

silviodonato:
This PR removes all 'analysis' categories which are shared with other groups

cc @cms-sw/analysis-l2 @cms-sw/reconstruction-l2 @cms-sw/xpog-l2


silviodonato:
merge

PR #1470, opened 2021-01-15
fix addon reporting

smuzaffar:


smuzaffar:
please test with cms-sw/cmssw#31719

smuzaffar:
please test with cms-sw/cmssw#31722

PR #1476, opened 2021-01-27
Init loop internal variable

mrodozov:
a test failed
https://cmssdt.cern.ch/jenkins/job/ib-run-pr-relvals/672/consoleFull
with:
`UnboundLocalError: local variable 'out_directory' referenced before assignment`
because the variable didn't exist outside the loop's scope. Maybe a condition exists in which the var is not set in that loop

### DIFF
--- a/report-pull-request-results.py
+++ b/report-pull-request-results.py
@@ -61,6 +61,7 @@ def parse_workflow_info( parts, relval_dir ):
   # it starts asumed as not found
   out_file = MATRIX_WORKFLOW_STEP_LOG_FILE_NOT_FOUND
   workflow_info[ 'step' ] = MATRIX_WORKFLOW_STEP_NA
+  out_directory = None
   for i in range( 0 , len( parts ) ):
     current_part = parts[ i ]
     if ( current_part == 'cd' ):


report-pull-request-results.py
```diff
@@ -61,6 +61,7 @@ def parse_workflow_info( parts, relval_dir ):
   # it starts asumed as not found
   out_file = MATRIX_WORKFLOW_STEP_LOG_FILE_NOT_FOUND
   workflow_info[ 'step' ] = MATRIX_WORKFLOW_STEP_NA
+  out_directory = None
```
smuzaffar, line 64: this alone is not going to fix the issue. If `out_directory` is None then later `join` statement will fail.

mrodozov: i'll try to get the conditions (arguments) of that function from the jenkins job then.

smuzaffar: no idea what you mean by that. Thsi function is reading the output of runTheMatrix and trying to find the failed workflow,  its directory path and step number. No idea how jenkins jobs is going to help you here

mrodozov: looks like this job didn't fail because of the function but because the relvals were failing (I'll guess das queries)

smuzaffar: job failed as it was expecting one-line message like `ERROR executing  cd 4.4_RunMinBias2012A+RunMinBias2012A+HLTD+RECODR1reHLT+HARVESTDR1reHLT; ...`  but it found message on multiple line
```
ERROR executing  
cd 4.34_ValSkim2011B+ValSkim2011B+HLTDSKIM+RECODR1reHLT+HARVESTDR1reHLT; 
```
this causes to not find `cd` in the first line

smuzaffar:
this change is not fixing the real issue. I suggest that as such failure are very low, so just re-build the failed job

mrodozov:
sure, reading what's the issue it seems the function is fine and when we want to fix it should be the input

PR #1480, opened 2021-02-02
 Update tag versions on non-default branches

mrodozov:
This updates non default branches given they has at least one manually created tag
The suggested usage in here https://github.com/cms-data/RecoBTag-Combined/issues/41 
doesn't work because when the PR is merged the last sha of the branch changes and doesn't match the
last tag on it, and the job is always first merging the PR and then attempt to tag the branch.
Tested for
default branch
non-default with tags - in most cases the non default is branched from the default and has some tags
non-default without any tags - orphan branch with 0 tags

### DIFF
--- a/create-new-data-pr.py
+++ b/create-new-data-pr.py
@@ -51,7 +51,7 @@ if __name__ == "__main__":
   data_pr_base_branch = data_repo_pr.base.ref
   data_repo_default_branch = data_repo.default_branch
   # create master just to exist on the cms-data repo if it doesn't
-  if data_repo_default_branch == "main":
+  if data_repo_default_branch != "master":
       if "master" not in [branch.name for branch in data_repo.get_branches()]:
           data_repo.create_git_ref(ref='refs/heads/master', sha=data_repo.get_branch(data_repo_default_branch).commit.sha)
 
--- a/create-new-data-pr.py
+++ b/create-new-data-pr.py
@@ -45,6 +45,9 @@ if __name__ == "__main__":
   data_prid = int(opts.pull_request)
   dist_repo = gh.get_repo(opts.dist_repo)
   data_repo_pr = data_repo.get_pull(data_prid)
+  if not data_repo_pr.merged:
+      print('The pull request isn\'t merged !')
+      exit(0)
   data_pr_base_branch = data_repo_pr.base.ref
   data_repo_default_branch = data_repo.default_branch
   # create master just to exist on the cms-data repo if it doesn't
@@ -52,28 +55,8 @@ if __name__ == "__main__":
       if "master" not in [branch.name for branch in data_repo.get_branches()]:
           data_repo.create_git_ref(ref='refs/heads/master', sha=data_repo.get_branch(data_repo_default_branch).commit.sha)
 
-  last_release_tag = None
-  repo_name_only = opts.data_repo.split("/")[1]
-  #find the last tag on the default branch, which is needed in all cases
-  err, out = run_cmd("rm -rf repo && git clone --bare https://github.com/%s -b %s repo && GIT_DIR=repo git log --pretty='%%d'" % (opts.data_repo, data_repo_default_branch))
+  err, out = run_cmd("rm -rf repo && git clone --bare https://github.com/%s -b %s repo && GIT_DIR=repo git log --pretty='%%d'" % (opts.data_repo, data_pr_base_branch))
   last_release_tag = get_tag_from_string(out)
-
-  if (data_pr_base_branch != data_repo_default_branch):
-      err, o = run_cmd("rm -rf repo && git clone --bare https://github.com/%s -b %s repo && GIT_DIR=repo git log --pretty='%%d'" % (opts.data_repo, data_pr_base_branch))
-      non_default_branch_last_tag = get_tag_from_string(o)
-      if not non_default_branch_last_tag:
-          print('Tags doesn\'t exist yet on %s branch, please create one manually first' % data_pr_base_branch)
-          exit(1)
-      elif (last_release_tag == non_default_branch_last_tag):
-          #the non default has the same tag as the default, non default was just branched out
-          print('Branch %s was just branched out' % data_pr_base_branch)
-          exit(1)
-      else:
-          last_release_tag = non_default_branch_last_tag
-
-  if not data_repo_pr.merged:
-      print('The pull request isn\'t merged !')
-      exit(0)
 
   if last_release_tag:
     comparison = data_repo.compare(data_pr_base_branch, last_release_tag)
--- a/create-new-data-pr.py
+++ b/create-new-data-pr.py
@@ -7,6 +7,7 @@ from os.path import expanduser
 from urllib2 import urlopen
 from json import loads
 import re
+from _py2with3compatibility import run_cmd
 
 def update_tag_version(current_version=None):
     updated_version = None
@@ -16,6 +17,15 @@ def update_tag_version(current_version=None):
             updated_version = '0'+updated_version
 
     return updated_version
+
+def get_tag_from_string(tag_string=None):
+    tag = None
+    for i in tag_string.split('\n'):
+        m = re.search("(V[0-9]{2}-[0-9]{2}-[0-9]{2})", i)
+        if m:
+           tag = m.group()
+           break
+    return tag
 
 if __name__ == "__main__":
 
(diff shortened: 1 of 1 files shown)


create-new-data-pr.py
```diff
@@ -35,29 +45,39 @@ def update_tag_version(current_version=None):
...
-          data_repo.create_git_ref(ref='refs/heads/master', sha=data_repo.get_branch(data_repo_base_branch).commit.sha)
+          data_repo.create_git_ref(ref='refs/heads/master', sha=data_repo.get_branch(data_repo_default_branch).commit.sha)
 
-  if (data_repo_base_branch != "master") and (data_repo_base_branch != "main"):
-      print("I do not know how to tag a non-master (non-main) branch %s" % data_repo_pr.base.ref)
-      exit(1)
+  last_release_tag = None
+  repo_name_only = opts.data_repo.split("/")[1]
+  #find the last tag on the default branch, which is needed in all cases
+  err, out = run_cmd("rm -rf %s; git clone git@github.com:%s && cd %s && git checkout %s && git log --pretty=\'%%s, %%d\' %s | grep tag:" % (repo_name_only, opts.data_repo, repo_name_only, data_repo_default_branch, data_repo_default_branch))
```
smuzaffar, line 58: better to use https://github.com and you can also simplify it 
```
"rm -rf repo && git clone --bare https://github.com/%s -b %s repo && GIT_DIR=repo git log --pretty='%%s, %%d'" % ( repo, branch)
```

create-new-data-pr.py
```diff
@@ -35,29 +45,39 @@ def update_tag_version(current_version=None):
...
       if "master" not in [branch.name for branch in data_repo.get_branches()]:
-          data_repo.create_git_ref(ref='refs/heads/master', sha=data_repo.get_branch(data_repo_base_branch).commit.sha)
+          data_repo.create_git_ref(ref='refs/heads/master', sha=data_repo.get_branch(data_repo_default_branch).commit.sha)
 
-  if (data_repo_base_branch != "master") and (data_repo_base_branch != "main"):
-      print("I do not know how to tag a non-master (non-main) branch %s" % data_repo_pr.base.ref)
-      exit(1)
+  last_release_tag = None
+  repo_name_only = opts.data_repo.split("/")[1]
+  #find the last tag on the default branch, which is needed in all cases
```
smuzaffar, line 57: @mrodozov , I do not understand this comment? why tag on the pr branch is not enough?

create-new-data-pr.py
```diff
@@ -89,6 +109,10 @@ def update_tag_version(current_version=None):
           print('files were modified, update mid version and reset minor', sec, thrd)
       new_tag = 'V'+first+'-'+sec+'-'+thrd
       # message should be referencing the PR that triggers this job
+      pr_branch_sha = data_repo.get_branch(data_pr_base_branch).commit.sha
+      #there is a bug in old versions, we need this three calls instead of one: https://github.com/PyGithub/PyGithub/issues/1336
+      gh_tag = data_repo.create_git_tag(tag=new_tag,message='Details in: '+data_repo_pr.html_url,object=pr_branch_sha,type='tag type')
```
smuzaffar, line 114: @mrodozov , just creating the tag is enough. There is no need to create ref and release

create-new-data-pr.py
mrodozov: in case we just branched out the default branch as another barnch, made PR on top of it and merged it the last tags of the non-default and the default branch will be the same, and need to be compared. 
Example is - we branch out 10_6_X from master, but now 10_6_X is the same as master. If we make a PR from 10_6_X and merge it, new tag will be created with greater number than the one existing on master, which shouldn't happen. So comparison with the default branch is always needed. Also if the branch is the default - the other branch doesn't go trough checks so it's not even redundant

create-new-data-pr.py
mrodozov: there is a bug, see the comment and the URL.
In the version we use create_git_tag it doesn't create the tag, I checked it. I took the example that worked from that github issue

create-new-data-pr.py
smuzaffar: We do not branch off from the tip of default. We keep on adding/merging on default branch. We only create a new branch when there are already changes in default which should not go in the non-default branch. So in practice we only create a new branch from an old tag.

create-new-data-pr.py
smuzaffar: ah ok, so create_git_tag only create a tag object ... do we need to create this tag objec? I think only calling
```
data_repo.create_git_ref(ref='refs/tags/'+new_tag, sha=pr_branch_sha)
```
is enough to create a tag

mrodozov: yes we need it's gh_tag.sha which is used to create the tag as a reference in 
`create_git_ref(ref='refs/tags/'+new_tag, sha=gh_tag.sha)`. so we only create it to get the tag sha from it which is later used to create the actual tag , meaning the  create_git_tag is only half working but just enough to make use of it.
The release is ... for human reading :D I can remove the release if we don't want it

smuzaffar: gh_tag.sha is same as pr_branch_sha so there is no need to create an tag object

mrodozov: it's not, this was my first thought the tag has it's own sha, not the one that it's tagging.
I'm checking it again but I'm sure (because I already did)

smuzaffar: On my repo https://github.com/smuzaffar/RecoBTag-Combined , the CMSSW_10_6_X branch sha is `aab813614cc883665f5ac86ff01c9a8e09741caa` ( https://api.github.com/repos/smuzaffar/RecoBTag-Combined/branches ) and the following works for me 
```
data_repo.create_git_ref(ref='refs/tags/X00-03-01', sha='aab813614cc883665f5ac86ff01c9a8e09741caa')
```

mrodozov: this is the output of this test
```
print("gh_tag.sha is:", gh_tag.sha, " pr_branch_sha is: ", pr_branch_sha) 
gh_tag.sha is: 929412f85052dd98168a85fe216ab2fb57d02243  pr_branch_sha is:  b473a61a7611abbadd45aa23fcc2dc21da1cc23d
```
the gh_tag.sha is the tag object sha, the pr_branch_sha is different

smuzaffar: this is what I get
```
>>> data_repo.get_branch('CMSSW_10_6_X').commit.sha
u'aab813614cc883665f5ac86ff01c9a8e09741caa'
```
so just using `data_repo.get_branch('CMSSW_10_6_X').commit.sha` as sha is good enogh to create the ref ( why make an extra API call? )

smuzaffar: we only need lightweight tags 

https://docs.github.com/en/rest/reference/git#tags
Note that creating a tag object does not create the reference that makes a tag in Git. If you want to create an annotated tag in Git, you have to do this call to create the tag object, and then create the refs/tags/[tag] reference. **If you want to create a lightweight tag, you only have to create the tag reference - this call would be unnecessary.**

mrodozov: I'm trying it

create-new-data-pr.py
```diff
@@ -35,29 +45,38 @@ def update_tag_version(current_version=None):
...
-  if (data_repo_base_branch != "master") and (data_repo_base_branch != "main"):
-      print("I do not know how to tag a non-master (non-main) branch %s" % data_repo_pr.base.ref)
-      exit(1)
+  last_release_tag = None
+  repo_name_only = opts.data_repo.split("/")[1]
+  #find the last tag on the default branch, which is needed in all cases
+  err, out = run_cmd("rm -rf repo && git clone --bare https://github.com/%s -b %s repo && GIT_DIR=repo git log --pretty='%%d'" % (opts.data_repo, data_repo_default_branch))
+  last_release_tag = get_tag_from_string(out)
+
+  if (data_pr_base_branch != data_repo_default_branch):
```
smuzaffar, line 61: remove this whole block and just use `data_pr_base_branch` in the above git clone command. No need to get latest tags for default branch

create-new-data-pr.py
mrodozov: it works you are right, only creating a reference with the pr commit tags it properly.

mrodozov: do we want the release or you want me to remove the call as well (as discussed being redundant)
sorry, I didn't notice this:
https://github.com/cms-sw/cms-bot/pull/1480/files#r569306705
in time

smuzaffar: deleyte the release creation call. For data repo we only need tag

mrodozov: it's deleted

create-new-data-pr.py
mrodozov: removed

create-new-data-pr.py
mrodozov: fixed as proposed

mrodozov:
Latest changes tested with bot branch on master and non-default-branch
see the last two tags here:
https://github.com/npc-data/PhysicsTools-NanoAOD/tags
and the last two PRs here:
https://github.com/npcbot/cmsdist/pulls

create-new-data-pr.py
```diff
@@ -35,29 +45,21 @@ def update_tag_version(current_version=None):
...
-      exit(1)
-
   if not data_repo_pr.merged:
-      print('Branch has not been merged !')
+      print('The pull request isn\'t merged !')
       exit(0)
+  data_pr_base_branch = data_repo_pr.base.ref
+  data_repo_default_branch = data_repo.default_branch
+  # create master just to exist on the cms-data repo if it doesn't
+  if data_repo_default_branch == "main":
```
smuzaffar, line 54: I think this condition should be 
```
if data_repo_default_branch != "master":
```

mrodozov: agree, it's better. its fixed

smuzaffar:
looks good.

PR #1482, opened 2021-02-08
add tau-pog watchers

swozniewski:
Defines tau-pog group and adds watch list.


smuzaffar:
any objections @mbluj

mbluj:
I confirm my agreement, please go on.

PR #1489, opened 2021-02-15
unset SINGULARITY_BIND inside singularity

smuzaffar:
as reported here https://github.com/cms-sw/cmssw/issues/32918

### DIFF
--- a/docker_launcher.sh
+++ b/docker_launcher.sh
@@ -117,7 +117,7 @@ if [ "X$DOCKER_IMG" != X -a "X$RUN_NATIVE" = "X" ]; then
       SINGULARITY_OPTIONS="${SINGULARITY_OPTIONS} -B $HOME:/home/cmsbuild"
     fi
     ERR=0
-    precmd="export SINGULARITY_BIND=''; "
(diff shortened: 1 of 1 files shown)


smuzaffar:
test parameters:
- addpkg = HeterogeneousCore/SonicTriton

smuzaffar:
please test

kpedro88:
@smuzaffar maybe it's worth copying the value of this variable into another variable that won't be recognized by Singularity, so that users can still check what was bound after activating the container?

kpedro88:
@smuzaffar actually, since this only affects the bot, maybe it doesn't matter. But we might want to consider whether this should be a default feature in https://github.com/cms-sw/cmssw-osenv/blob/master/cmssw-env

smuzaffar:
ah I notcied that instead singularity the variable is `SINGULARITY_BIND` while singularity uses `SINGULARITY_BINDPATH` for binding , so may be this is not needed

smuzaffar:
singularity docs  ( https://sylabs.io/guides/3.0/user-guide/bind_paths_and_mounts.html ) says that it uses both `SINGILARITY_BIND` and `SINGILARITY_BINDPATH`

smuzaffar:
please test

smuzaffar:
please test

PR #1493, opened 2021-02-24
Replace Javier with Cecile as l1t manager

silviodonato:
@jmduarte @cecilecaillol @rekovic 
Could you confirm everything is correct?


smuzaffar:
any objections @cecilecaillol

cecilecaillol:
> any objections @cecilecaillol

No that sounds good

PR #1495, opened 2021-02-28
Create empty eventsetuprecord-get.txt

gartung:
Older CMSSW releases do not generate eventsetuprecord-get.txt.


gartung:
@smuzaffar

PR #1497, opened 2021-03-03
Enable multiple workflows for IB profiling job.

gartung:


schedule-additional-tests
```diff
@@ -30,9 +30,12 @@ for x in `grep ADDITIONAL_TESTS=  $CONFIG_MAP | grep "SCRAM_ARCH=$ARCHITECTURE;"
   fi
   if [ "$x" = "profiling" ] ; then
     PROFILING_WORKFLOWS=$(echo $(grep "PR_TEST_MATRIX_EXTRAS_PROFILING=" $CMS_BOT_DIR/cmssw-pr-test-config | sed 's|.*=||'), | tr ' ' ','| tr ',' '\n' | grep '^[0-9]' | sort | uniq | tr '\n' ',' | sed 's|,*$||')
```
smuzaffar, line 32: also change the above line to
```
PROFILING_WORKFLOWS=$(grep "PR_TEST_MATRIX_EXTRAS_PROFILING=" $CMS_BOT_DIR/cmssw-pr-test-config | sed 's|.*=||' | tr ',' ' ')
```

schedule-additional-tests
```diff
@@ -26,5 +26,15 @@ for x in `grep ADDITIONAL_TESTS=  $CONFIG_MAP | grep "SCRAM_ARCH=$ARCHITECTURE;"
     echo "EVENTS=10" >> ${nfile}
     rm -f ${pfile}
   fi
+  if [ "$x" = "profiling" ] ; then
+    PROFILING_WORKFLOWS=$(grep "PR_TEST_MATRIX_EXTRAS_PROFILING=" $CMS_BOT_DIR/cmssw-pr-test-config | sed 's|.*=||' | tr ',' ' ')
+    for wf in $PROFILING_WORKFLOWS; do
+      nfile=$WORKSPACE/auto-$x-${wf}-properties
+      cp ${pfile} ${nfile}
+      echo "PROFILING_WORKFLOWS=$PROFILING_WORKFLOWS" >> ${pfile}
```
smuzaffar: you should use `${wf}` instead of `$PROFILING_WORKFLOWS` here. Also use `${nfile}` . You are also missing 
```
echo "PROFILE=pp" >> ${nfile}
```

gartung: I changed the parameters of the job to not use PROFILE or WORKFLOWS.

schedule-additional-tests
```diff
@@ -29,10 +29,13 @@ for x in `grep ADDITIONAL_TESTS=  $CONFIG_MAP | grep "SCRAM_ARCH=$ARCHITECTURE;"
     rm -f ${pfile}
   fi
   if [ "$x" = "profiling" ] ; then
-    PROFILING_WORKFLOWS=$(echo $(grep "PR_TEST_MATRIX_EXTRAS_PROFILING=" $CMS_BOT_DIR/cmssw-pr-test-config | sed 's|.*=||'), | tr ' ' ','| tr ',' '\n' | grep '^[0-9]' | sort | uniq | tr '\n' ',' | sed 's|,*$||')
-    prof="pp"
-    echo "PROFILE=$prof" >> ${pfile}
```
smuzaffar, line 34: @gartung you are missing this in new file

smuzaffar: you can have something like
```
echo "PROFILE=$prof" >> ${pfile}
echo "EVENTS=20" >> ${pfile}
```

in $pfile as these are common variables and only add  workflow number for $nfile

gartung: I removed that parameter since it has no meaning. Both pp and mp profiles are run by the profiling scripts.

smuzaffar: ah ok

smuzaffar:
@gartung , looks good. Is it ready for merge?

gartung:
The 11834.21 workflow is crashing in step2 so I need to remove it.

smuzaffar:
@gartung , currently the profiling job runs on normal build nodes ( which are 16 core vms) and it waste a lot of resources. I have created two dedicated 4 cores nodes and update the job to use those nodes .

gartung:
OK.

gartung:
@smuzaffar Can this be merged. You checked in your version of cmssw-pr-test-config with workflows 2 and 3 (for testing?). The profiling jobs fail because they don't exist.

gartung:
https://cmssdt.cern.ch/jenkins/job/ib-run-profiling/676/parameters/

PR #1498, opened 2021-03-04
Split profiling workflows the same way for igprof and profiling jobs. 

gartung:
Each job gets its own properties file. The run-ib-igprof script was also updated to split the WORKFLOWS var for use in a loop.


gartung:
@smuzaffar Please merge this.

PR #1500, opened 2021-03-08
reco comparisons: add a set of mappings for D76 similar to existing D49

slava77:
covers workflows 34634.0,34834.0,34834.99,34834.999,34834.21

this should resolve warnings ( `comparisons for the following workflows were not done due to missing matrix map` ) appearing since Mar 5.


slava77:
@cmsbuild please test

slava77:
the update worked as expected based on https://github.com/cms-sw/cms-bot/pull/1500#issuecomment-793253423.
So, I think that it can be merged.

smuzaffar:
+externals

PR #1501, opened 2021-03-10
copy igprof mem interim outputs

jpata:
The current igprof MEM analysis copied the final output file specified with the -o option. As we learned, this should not be used, as it contains incorrect or misleading data (the MEM usage is much lower than what is expected).

Instead, we want to make sure we copy IgProf.N.gz files created by the customize function, and the IgProf.99.gz file (for N=100) should be used for the final MEM analysis.

AFAIK, using the igprof pp / CPU file specified by -o is still fine.

cc @slava77


jpata:
I submitted a test job for this by hand here: https://cmssdt.cern.ch/jenkins/job/release-run-reco-profiling/51/

slava77:
> Instead, we want to make sure we copy IgProf.N.gz files created by the customize function, and the IgProf.99.gz file (for N=100) should be used for the final MEM analysis.

in https://jiwoong.web.cern.ch/jiwoong/ we were careful to provide all 3 mem files. I think that we need at least 2, .1 and .99 (or N-1 if N is not 100).

jpata:
Yes, we copy all now: https://github.com/cms-sw/cms-bot/pull/1501/files#diff-b1c5281a2029a6fc2d815017134a2192e32b7221b725e2e60e0eb46beb124bbcR139-R147

jpata:
rerunning after a fix (different `/` behaviour with different python versions): https://cmssdt.cern.ch/jenkins/job/release-run-reco-profiling/53

jpata:
The job succeeded and the MEM igprof output is now fixed when parsing the file `step3_igprofMEM.99.gz`: https://jpata.web.cern.ch/jpata/cgi-bin/igprof-navigator/releases/11_3_0_pre4/23434.21/step3/mem/

@smuzaffar, good to merge from my side, thanks!

slava77:
> The job succeeded and the MEM igprof output is now fixed when parsing the file `step3_igprofMEM.99.gz`: https://cms-reco-profiling.web.cern.ch/cms-reco-profiling/cgi-bin/igprof-navigator/releases/CMSSW_11_3_0_pre4/23434.21/step3/mem/16

I think that we need individual reports per case.

I guess this is a matter of updating the parsing and copy to https://cms-reco-profiling.web.cern.ch in some other repo.

jpata:
All the .1.gz, .50.gz and .99.gz are copied to EOS at the end of profiling, it's a matter of Bongho's / my scripts to parse them to sql. Currently, my interim scripts only parse .99, but I think the "official" web page can contain the outputs from Bongho.

PR #1505, opened 2021-03-14
adding PR checks on edm::TriggerResults

missirol:
This PR is an attempt to include automatic comparisons for the `edm::TriggerResults` collections in the outputs produced during PR tests in [cms-sw/cmssw](https://github.com/cms-sw/cmssw).

It was recently mentioned by TSG experts that such additional checks could be useful to validate the HLT results in PR reviews.

Even though the original use case is related to the HLT outputs, the implementation is fairly generic, in that it checks all the `edm::TriggerResults` collections in a given workflow (without choices tailored towards HLT).

The PR adds a python executable, `compareTriggerResults`, which is basically a wrapper calling the CMSSW executable `hltDiff` on the relevant inputs, and producing an output directory with the relevant results. The output directory includes a file with a summary of the comparisons' results, in `html` or text format.

The PR also includes an update of `run-pr-comparisons`, in order to run these additional comparisons. The outputs of these new checks are saved in a subfolder named `triggerResults/`, created in the same directory as (for example) the subfolder `validateJR/`. A 1-line summary of these TriggerResults comparisons is also added to the file `qaResultsSummary.log`, which (afaiu) is what provides the summary message of automatic tests in [cms-sw/cmssw](https://github.com/cms-sw/cmssw) PR reviews.

Local tests suggest that the new comparisons take roughly 1min per workflow. If this is too long, maybe these checks could be made optional. This would require feedback from experts (for the moment, I left a placeholder flag for this [here](https://github.com/cms-sw/cms-bot/compare/master...missirol:master#diff-7f111d1913bc7caf04aaa0bff3ba37def0e16e8efa3aebe20d449602c54e524aR328)).

Attn: @fwyzard @Martin-Grunewald @Sam-Harper

--------

**PR Validation:**

Tested `compareTriggerResults` locally with copies of EDM outputs of 2 workflows, from recent CMSSW PRs (could not test locally the updates to run-pr-comparisons).


fwyzard:
please test

fwyzard:
actually, can the bot run tests on itself ?

smuzaffar:
> actually, can the bot run tests on itself ?

yes

missirol:
latest commits address the following points:
* paths to input directories were incorrect in my additions to `run-pr-comparisons` (this explains the errors)
* enforced that additions to the PR summary message correspond to (at most) 1 line

smuzaffar:
please test

missirol:
As far as I can see, it works as intended (see the "TriggerResults" link in the Comparison Summary). I noticed that the workflows were not ordered correctly in the relevant table, so I tried to fix that in the last commit.

smuzaffar:
@missirol , so this additional test is taking nearly 30mins to run. Can we run it in parallel e.g. `compareTriggerResults` can be run for each workflow separatly and then at the end you can run something to summarize the results? See https://github.com/cms-sw/cms-bot/blob/ad7dcea21d8b937dd4f9d2cab03f96bcfeac5588/run-pr-comparisons#L196-L226 where we do it for other comparison jobs . All you need is to 
```
for each wf
  wait if already more than  ${NUM_PROC} jobs running
  run compareTriggerResults with/ref/wf with/test/wf & 
done
wait for all jobs to be done
generate summary
```

also move your code ( https://github.com/cms-sw/cms-bot/blob/ad7dcea21d8b937dd4f9d2cab03f96bcfeac5588/run-pr-comparisons#L327-L336 ) out side the `if [ "X$RUN_JR_COMP" = Xtrue ]; then` condition

missirol:
Thanks for reviewing, @smuzaffar .

> Can we run it in parallel e.g. `compareTriggerResults` can be run for each workflow separatly and then at the end you can run something to summarize the results?

Okay, I will give this a try (it might take me a few days to update this PR).

missirol:
The latest updates should allow to run the new comparisons in parallel, as suggested [here](https://github.com/cms-sw/cms-bot/pull/1505#issuecomment-799224537).
A new script (`compareTriggerResultsSummary`) is added to harvest the results, and produce summary information.

fwyzard:
please test

smuzaffar:
thanks @missirol , it looks good now (only took 3 mins to run the extra comparison test). 
@fwyzard , if this looks good to you then I am ready to merge it.

missirol:
Hi @smuzaffar , thanks again for your suggestion. Just for my education, where can you check how long the new comparisons took?

fwyzard:
>  @fwyzard , if this looks good to you then I am ready to merge it.

Looks good to me.
I guess we'll need to see how it behave when some real changes in the TriggerResults happen :)

smuzaffar:
You can click on the "Show all checks" link at the end of this PR page and then click on the comparison status "Details" link which should take you to the jenkins job for comparison ( https://cmssdt.cern.ch/jenkins/job/compare-root-files-short-matrix/41714/ ) . There you go to the "view as plain text" and search for "edm::TriggerResults" to see the time stamps e.g. in this case I see
```
[Tue Mar 16 01:25:42 CET 2021] Starting comparisons of edm::TriggerResults
...
[Tue Mar 16 01:27:39 CET 2021] Running comparisons of edm::TriggerResults ...
[Tue Mar 16 01:28:44 CET 2021] Finished comparisons of edm::TriggerResults
```

missirol:
okay, thanks!

smuzaffar:
@missirol , @fwyzard any idea how this will behave for old release cycles e.g. CMSSW 5.3.X to 11.2.X ?

fwyzard:
Looking at the git logs, `hltDiff` has been fairly stable since CMSSW 7.5.x until Marino's fixes.

Can you enable these checks only in CMSSW_8_0_X IBs and later ?

fwyzard:
Let me rephrase...

@missirol do you expect the check to work also without your recent fixes ?
If so:
  - @smuzaffar can you enable the new check only for CMSSW_8_0_X and later IBs ?

Otherwise, if the fixes are needed for the checks to work:
  - @missirol could you make a couple of backprt PRs for CMSSW 11.1.x and 11.2.x ?
  - @smuzaffar can you enable the new check only for CMSSW_11_1_X and later IBs ?

missirol:
Hi @fwyzard , (by fixes, I think we mean cms-sw/cmssw#33174)

I don't think they need to be backported, because those fixes only had to do with badly mis-configured cmd-line arguments. For example, in these PR checks, the process name will be correct by definition, and the output directories are created appropriately before `hltDiff` is used. cms-sw/cmssw#33174 really just fixed a few fringe cases that should never show up here.

Obviously, it's difficult to be certain that all this will work in earlier releases, but (as you wrote) `hltDiff` didn't change drastically in recent times, and in principle the python scripts added here have some checks to handle ill-defined inputs, plus the `|| true` calls in `run-pr-comparion` should ensure that PR checks aren't stopped by issues with these updates.

smuzaffar:
As @missirol as properly protected the commands with `|| true` so comparison job itself will not fail. Anyway, I am opening a dummy cms-bot PR to run tests all release cycles

smuzaffar:
see https://github.com/cms-sw/cms-bot/pull/1508

PR #1519, opened 2021-03-24
add HCALOnly wf to MATRIX_EXTRAS_GPU

slava77:
it looks like we are not testing the HCAL GPU workflow.

@fwyzard  was it intentional (that .522 is broken/unstable) or a result of minimization of the number of workflows?


fwyzard:
Not intentional, I guess it simply for lost ?

fwyzard:
enable gpu

fwyzard:
please test

slava77:
@smuzaffar @mrodozov 
is there a way to make baseline IB outputs for MATRIX_EXTRAS_GPU?

smuzaffar:
@slava77 , no currently there is no option

slava77:
> @slava77 , no currently there is no option

it would be nice to get it added some time soon.
Well, by "it" I include also a case when these (now 3) workflows run for every IB.
Some more optimal variant could be to trigger the baseline when missing for cases when there is a PR which has "enable gpu" as a part of "please test", but could also be a manually "please check baseline" based on the PR test parameters so that we can have a "N-1" baseline possibility as well.

For now, I think that this PR is OK considering that all 3 workflows completed.

smuzaffar:
@slava77 , #1521 should enable GPU comparisons

PR #1522, opened 2021-03-26
reco comparisons: add maps for pixel/ecal/hcal-only CPU and GPU workflows

slava77:


slava77:
enable gpu

slava77:
@cmsbuild please test

smuzaffar:
+externals
looks good

PR #1529, opened 2021-04-01
always add extra merge commit even if there are no cmssw PRs involved

smuzaffar:


smuzaffar:
test parameters:
- disable_poison = true

smuzaffar:
please test with cms-sw/cmsdist#6777,cms-sw/root#153

smuzaffar:
@slava77 , as mentioned https://github.com/cms-sw/root/pull/153#issuecomment-811862603, bot was only checking for extra merges when cmssw PRs are involved. 
For non-cmssw PRs , we do not get extra merge commits. In this case bot just runs "git cms-addpkg packages` which checks out packages from the release tag (instead of release branch). So for non-cmssw PR tests there will be no extra merges checkout.

So, I think this change is not needed.

slava77:
> @slava77 , as mentioned [cms-sw/root#153 (comment)](https://github.com/cms-sw/root/pull/153#issuecomment-811862603), bot was only checking for extra merges when cmssw PRs are involved.
> For non-cmssw PRs , we do not get extra merge commits. In this case bot just runs "git cms-addpkg packages` which checks out packages from the release tag (instead of release branch). So for non-cmssw PR tests there will be no extra merges checkout.
> 
> So, I think this change is not needed.

I'm not sure I followed the description above.
The symptom in cms-sw/root#153 was clear that a test was done with the head of cms-sw/cmsdist (not equal to the baseline IB).
Is your comment implying that there is no way to see the extra cms-sw/cmsdist merges in the test summary? (if I can do it manually, I'd guess that it can be done programmatically).

smuzaffar:
No, I was only talking about cmssw as currently bot is only looking for cmssw extra merges. We should be able to get the extra merges for cmsdist too.

PR #1533, opened 2021-04-14
Update watchers.yaml for ecal-offline

thomreis:
Extend watchers package list for ecal-offline.


mrodozov:
please test
lets check the bot is happy I don't see typos but still

PR #1538, opened 2021-04-19
Only run the cms-esrget clang-tidy check on pull requests.

gartung:
To avoid confusion only run the cms-esrget clang-tidy check during pull request test job.
The cms-handle check offers fixit text that compiles 100% the last time it was checked for all of CMSSW and should be run as part of the code-checks job.

### DIFF
--- a/pr_testing/test_multiple_prs.sh
+++ b/pr_testing/test_multiple_prs.sh
@@ -719,7 +719,7 @@ if [ "X$DO_STATIC_CHECKS" = "Xtrue" -a "X$CMSSW_PR" != X -a "$RUN_TESTS" = "true
     curl -s -L https://patch-diff.githubusercontent.com/raw/${PR_REPO}/pull/${PR_NUMBER}.patch | grep '^diff --git ' | sed 's|.* a/||;s|  *b/.*||' | sort | uniq > $WORKSPACE/all-changed-files.txt
     cat $WORKSPACE/all-changed-files.txt
     echo ""
-    USER_CXXFLAGS='-Wno-register -DEDM_ML_DEBUG -w' SCRAM_IGNORE_PACKAGES="Fireworks/% Utilities/StaticAnalyzers" USER_CODE_CHECKS='cms*' scram b -k -j ${NCPU2} code-checks USER_CODE_CHECKS_FILE="$WORKSPACE/all-changed-files.txt" SCRAM_IGNORE_SUBDIRS=test 2>&1 | tee -a $WORKSPACE/llvm-analysis/runCMSClangTidyChecks.log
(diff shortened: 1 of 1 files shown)


gartung:
@makortel

gartung:
please test

gartung:
abort

gartung:
abort

gartung:
Passed when tested with https://github.com/cms-sw/cmssw/pull/33480

makortel:
Based on the tests in cms-sw/cmssw#33480 it seems that the `getByToken` check would be already filtered out even without this PR?

gartung:
There are getByToken calls in the source file
https://github.com/cms-sw/cmssw/blob/c6c9b0fbbc0cfd3627f0185c9589d49acf9f6354/FastSimulation/Tracking/plugins/PixelTracksProducer.cc#L70

makortel:
> There are getByToken calls in the source file
> https://github.com/cms-sw/cmssw/blob/c6c9b0fbbc0cfd3627f0185c9589d49acf9f6354/FastSimulation/Tracking/plugins/PixelTracksProducer.cc#L70

Right, but the clang-tidy warnings shown in the last test https://github.com/cms-sw/cmssw/pull/33480#issuecomment-824178290 do not show any getByToken warnings.

gartung:
Because only the cms-esrget clang-tidy check is enabled.

makortel:
Right, but the last test https://github.com/cms-sw/cmssw/pull/33480#issuecomment-824178290 was without this PR, which made me wonder if we are already in the situation that only `cms-esrget` test is enabled (without this PR; and the period where the getByToken check was visible was short).

gartung:
I had 'cms-*' enabled before but only reported errors if there were cms-esrget warnings. The cms-handle warnings appear in the clang tidy log without this change
https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-c707a2/14384/llvm-analysis/runCMSClangTidyChecks.log

gartung:
Compare to https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-c707a2/14360/llvm-analysis/runCMSClangTidyChecks.log 
where only cms-esrget is enabled.

makortel:
Thanks, now I see, they were in the "clang-tidy log", not in the list of warnings. This PR should be good to go then.

mrodozov:
+externals

gartung:
abort

gartung:
please test

mrodozov:
this is confusing the bot, somehow getting also 33480

mrodozov:
-externals

mrodozov:
it's running relvals, it's fine

PR #1545, opened 2021-04-29
Allow profiling workflow(s) to be specified with PR multiline comment 

gartung:
Adding multiline comment
```
test parameters:
workflows_profiling=11834.21
```
should override the default 
`PR_TEST_MATRIX_EXTRAS_PROFILING=23434.21,11834.21`


gartung:
@smuzaffar

process_pr.py
```diff
@@ -55,6 +55,7 @@ def format(s, **kwds): return s % kwds
 EXTRA_TESTS = "gpu|threading|profiling|none"
 MULTILINE_COMMENTS_MAP = {
               "workflow(s|)(_gpu|_threading|)":  [format('^%(workflow)s(\s*,\s*%(workflow)s|)*$', workflow= WF_PATTERN),       "MATRIX_EXTRAS"],
+              "workflow(s|)(_profiling|)":  [format('^%(workflow)s(\s*,\s*%(workflow)s|)*$', workflow= WF_PATTERN),       "PROFILING_WORKFLOWS"],
```
smuzaffar, line 58: change it to `workflow(s|)_profiling` otherwise `workflows = num` will set PROFILING_WORKFLOWS instead of MATRIX_EXTRAS

schedule-additional-tests
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash -ex
 CONFIG_MAP=$(dirname $0)/config.map
-PROFILING_WORKFLOWS=$(grep "PR_TEST_MATRIX_EXTRAS_PROFILING=" $CMS_BOT_DIR/cmssw-pr-test-config | sed 's|.*=||' | tr ',' ' ')
+if [ "X$PROFILING_WORKFLOWS" == "X" ];then
+  PROFILING_WORKFLOWS=$(grep "PR_TEST_MATRIX_EXTRAS_PROFILING=" $CMS_BOT_DIR/cmssw-pr-test-config | sed 's|.*=||' | tr ',' ' ')
+fi
```
smuzaffar, line 5: this is only used for IBs test. So I do not think we need this change. For IBs we should get workflows from the cmssw-pr-test-config file. You need update https://github.com/cms-sw/cms-bot/blob/master/pr-schedule-tests#L102 and add a line 
```
echo "PROFILING_WORKFLOWS=${PROFILING_WORKFLOWS}" >> $OUTPUT_FILE
```

slava77:
should this PR still be restricted to just what's possible/tested in the context of https://github.com/cms-cmpwg/profiling/blob/master/Gen_tool/Gen.sh and what follow from it (4 steps etc) ?

I'm guessing that I can't ask for some random wf, like 136.88811 which we often use for data PAT/miniAOD testing.

smuzaffar:
@slava77 , yes you can not ask for any random workflow. Workflow should the one defined here https://github.com/cms-sw/cms-bot/blob/master/cmssw-pr-test-config#L3

smuzaffar:
@gartung , can we trigger an IB profiling job if PR profiling workflow is not part of  https://github.com/cms-sw/cms-bot/blob/master/cmssw-pr-test-config#L3 ? I think from the pr-schedule-tests we can check this and ask jenkins to start such job

gartung:
The profiling script expects a 4 step simulation workflow with simulation, pileup mixing, reconstruction, and miniAOD production. I am trying to figure out a way to validate the workflow against PR_TEST_MATRIX_EXTRAS_PROFILING

pr_testing/run-pr-profiling.sh
```diff
@@ -1,14 +1,23 @@
 #!/bin/bash -ex
 source $(dirname $0)/setup-pr-test-env.sh
 
-PROFILING_WORKFLOWS=$(grep "PR_TEST_MATRIX_EXTRAS_PROFILING=" $CMS_BOT_DIR/cmssw-pr-test-config | sed 's|.*=||;s|,| |')
+ALLOWED_PROFILING_WORKFLOWS=$(grep "PR_TEST_MATRIX_EXTRAS_PROFILING=" $CMS_BOT_DIR/cmssw-pr-test-config | sed 's|.*=||;s|,| |')
+
+for PROFILING_WORKFLOW in $PROFILING_WORKFLOWS;do
```
smuzaffar, line 6: please set `PROFILING_WORKFLOWS=${ALLOWED_PROFILING_WORKFLOWS}` if `$PROFILING_WORKFLOWS` is not set

gartung:
enable profiling

gartung:
test parameters:
workflows_profiling=11834.21

gartung:
please test

gartung:
abort

smuzaffar:
@gartung  There is no way we can test this as it is. Either we merge it or make two PRs  one for cms-bot so that it can properly pass the new profiling parameter and other to use it in profiling job.
I suggest to just merge it :-)

gartung:
OK Thanks.

PR #1564, opened 2021-06-10
Full re-build for cmssw PRs

smuzaffar:
work around git-cms-addpkg issue with added '*' if there are already packages checkedout


smuzaffar:
+externals
looks good

PR #1566, opened 2021-06-18
Substitute PPS software person

civanch:
It is requested by Fabrizio Ferro.


smuzaffar:
any objections @forthommel , @fabferro ? This change means that @fabferro (instead of @forthommel ) is going to receive github notifications emails for PPS packages?

forthommel:
No objections at all :) I would have simply copy-pasted my set as a duplication for @fabferro but if we are mutually exclusive... :)

fabferro:
Ok for me

PR #1569, opened 2021-06-22
Add lines to PR testing summary for Static Analyzers checks for ESR::get and legacy modules usage.

gartung:
Resolves https://github.com/makortel/framework/issues/99


gartung:
@makortel

makortel:
Thanks! Hmm, I suppose the best way to test would be to open test PRs touching such EDModules.

gartung:
please test with https://github.com/cms-sw/cmssw/pull/34213

gartung:
Appears to be doing what I expected
https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-2e7d67/16170/summary.html

gartung:
abort

smuzaffar:
please test with cms-sw/cmssw#34213
let it run, I want to see the final comment

gartung:
OK. I triggered the test on the test PR since the primary PR needs to be in CMSSW for the code-checks to work.

smuzaffar:
+externals

makortel:
The PR test message https://github.com/cms-sw/cmssw/pull/34213#issuecomment-866315629 looks otherwise very good, but I'm a bit afraid that the duplicated warnings about the `EventSetupRecord::get()` can be confusing (one set of warnings from static analyzer, another set from clang-tidy).

@gartung I recall you noticed that initially the clang-tidy checker found more cases than the static analysis check. Is this still the case?

gartung:
No. They are equivalent now.

makortel:
Thanks. Could we then show only one of the static analyzer / clang-tidy reports? To me the static analyzer warning looks more explicit in the PR message, and the `esrget-sa.txt` has a bit more details than `cmsclangtidy.txt`.

gartung:
OK. I will submit a PR that removes the clang-tidy warnings.

PR #1581, opened 2021-07-08
for full cmssw avoid running static checks

smuzaffar:


smuzaffar:
please test with cms-sw/cmssw#34396,

PR #1586, opened 2021-07-13
do not send buildlog to jenkins and keep on building

smuzaffar:


smuzaffar:
please test with cms-sw/cmssw#34453

PR #1594, opened 2021-07-19
reco comparisons: add a map entry for 11634.912 DD4hepDB workflow

slava77:
this should clear the warning that started to appear in tests since CMSSW_12_0_X_2021-07-18-2300


slava77:
@cmsbuild please test

smuzaffar:
+externals

PR #1599, opened 2021-07-27
Adding Andrea P. to operations

silviodonato:


silviodonato:
cc @perrotta

perrotta:
Thank you @silviodonato 
After our chat I already updated the file myself: I was expecting that it would have opened a PR (like this one), but I was unaware of the "superpower"s I got recently and the net result is that the change was automatically merged: here the conflict.

PR #1600, opened 2021-07-28
plots for ticlTrackstersHFNose* products 

slava77:
(these are currently present only in special HFNose workflows; one in the short matrix)

there is no coverage for these in the DQM; so this monitoring would be a good addition.

This PR is also in sync with cms-sw/cmssw#34605


slava77:
@cmsbuild please test

PR #1622, opened 2021-09-08
allow to use cmssw-config sources

smuzaffar:


smuzaffar:
please test with cms-sw/cmssw-config#86

PR #1636, opened 2021-09-21
Make a ubsan_logs link on the IB page

mrodozov:
not sure if this isn't going to put a link for every IB (not only for the ubsan ones)


smuzaffar:
@mrodozov , I do not think we are using this any more. please update https://github.com/cms-sw/cms-bot/blob/master/report-summary-merged-prs

PR #1638, opened 2021-09-23
Skip search for ubsan_ib-->non_ubsan

mrodozov:
the first case in this json where compared_tags is "compared_tags": "CMSSW_12_1_UBSAN_X_2021-09-22-1600-->CMSSW_12_1_X", 
https://raw.githubusercontent.com/cms-sw/cms-sw.github.io/master/data/CMSSW_12_1_UBSAN_X.json
should be skipped because it resolves to searching for CMSSW_12_1_X in ubsan_logs which will never be there
the release folder to be searched in has to be always the second  element of `rel_name = c['compared_tags'].split('-->')[1]`
and in this case will search with `find` into never existing path

### DIFF
--- a/report-summary-merged-prs
+++ b/report-summary-merged-prs
@@ -1068,7 +1068,7 @@ def find_check_headers(comparisons, architecture):
 
 def find_ubsan_logs(comparisons, arch):
     for c in comparisons:
-        if c['compared_tags'].find('UBSAN') < 2: continue
+        if c['compared_tags'].count('UBSAN') < 2: continue
         rel_name = c['compared_tags'].split('-->')[1]
         print('Looking for ubsan results for', rel_name, '.')
         c['ubsan-logs'] = find_and_check_result(rel_name, arch, CHECK_UBSANLOG_PATH, 'wc -l {0} | cut -d " " -f1')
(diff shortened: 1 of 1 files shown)


report-summary-merged-prs
```diff
@@ -1068,6 +1068,7 @@ def find_check_headers(comparisons, architecture):
 
 def find_ubsan_logs(comparisons, arch):
     for c in comparisons:
+        if c['compared_tags'].find('UBSAN') < 2: continue
```
smuzaffar, line 1071: what done `find()` return?

smuzaffar: I think you can just check of `UBSAN` in c['release_name']

mrodozov: integer with how many times the substring is found into the searchstring

mrodozov: sorry im thinking of count ill change it

smuzaffar: for me it returns the location where it find ubsan e.g
```
>python -c 'print("CMSSW_12_1_UBSAN_X_2021-09-22-1600-->CMSSW_12_1_X".find("UBSAN"))'
11
```

smuzaffar: I would suggest to just use `c['release_name']`

mrodozov: yes sorry it returns the position of the first hit , should be count not find

PR #1658, opened 2021-11-03
Add new package for monitoring pixel SoA products

sroychow:
This PR is with ref to CMSSW PR #[35965](https://github.com/cms-sw/cmssw/pull/35965) for introducing the new package - ```DQM/SiPixelPhase1Heterogeneous```


mmusich:
I am wondering if the @cms-sw/heterogeneous-l2 category should be assigned as well (but maybe not).

makortel:
If the aim of `DQM/SiPixelPhase1Heterogeneous` is only to monitor SoA's that live in the host memory and there would be no GPU code per se, I'd prefer to leave `heterogeneous` out.

fwyzard:
OK for me to leave `heterogeneous` out, at least for the time being.

PR #1659, opened 2021-11-04
reco comparisons: add DNN ID plots for electrons and photons

slava77:
To add supporting plots for ID variables added in cms-sw/cmssw#35403


slava77:
@cmsbuild please test

slava77:
the logs look OK (reconfirming already what I've seen in local tests).

It may be more practical to merge this soon after jenkins moves to CMSSW_12_2_X_2021-11-04-2300, which has cms-sw/cmssw#35984; prior to that this PR will show occasional random differences in two variables that are being read from uninitialized memory (out of range)

smuzaffar:
+external
Thanks @slava77

PR #1682, opened 2021-12-22
Access MariaDB with no password; use consistent unit test db name

amaltaro:
Starting with MariaDB 10.6.5, we no longer define a user password.
It also changes the MariaDB database name to the same defined within WMCore unittests, to avoid inconsistencies and surprises when testing things.

Further info can be found at: https://github.com/dmwm/WMCore/issues/10926


amaltaro:
This change has been tested in Jenkins and everything is working properly. @smuzaffar could you please merge it? Thanks

PR #1685, opened 2022-01-07
reco comparisons: add plots for new MTD value maps; add basic monitoring of JECs in pat::Jets

slava77:
- add plots for new MTD value maps following updates in cms-sw/cmssw#36202
- pat::Jet JEC counters in connection with cms-sw/cmssw#36559

I will test in cms-sw/cmssw#36559


slava77:
based on tests in  cms-sw/cmssw#36559 I see that the logs are showing that the new variables are picked up in the plots (there are no differences in that PR though).

So, I think that this is ready to be merged

perrotta:
+1

smuzaffar:
+externals

PR #1689, opened 2022-01-14
Move to slc7_amd64_gcc530 as production arch for CMSSW_8_0_36_patch1

smuzaffar:
see https://github.com/cms-sw/cmssw/issues/36708
FYI @qliphy @perrotta @cms-sw/pdmv-l2

### DIFF
--- a/releases.map
+++ b/releases.map
@@ -3204,5 +3204,5 @@ architecture=cc8_amd64_gcc9;label=CMSSW_12_2_0;type=Production;state=Announced;p
 architecture=slc7_aarch64_gcc9;label=CMSSW_12_2_0;type=Production;state=Announced;prodarch=0;
 architecture=slc7_amd64_gcc900;label=CMSSW_12_2_0;type=Production;state=Announced;prodarch=1;
 architecture=slc7_ppc64le_gcc9;label=CMSSW_12_2_0;type=Production;state=Announced;prodarch=0;
-architecture=slc6_amd64_gcc530;label=CMSSW_8_0_36_patch1;type=Production;state=Announced;prodarch=1;
-architecture=slc7_amd64_gcc530;label=CMSSW_8_0_36_patch1;type=Production;state=Announced;prodarch=0;
+architecture=slc6_amd64_gcc530;label=CMSSW_8_0_36_patch1;type=Production;state=Announced;prodarch=0;
+architecture=slc7_amd64_gcc530;label=CMSSW_8_0_36_patch1;type=Production;state=Announced;prodarch=1;


smuzaffar:
By the way, I see that `CMSSW_8_0_34_UL` was build for slc7_amd64_gcc530. Are we sure that we want CMSSW_8_0_36+patch or CMSSW_8_0_34_UL + patch

jordan-martins:
> 

hi guys, I would say that this is the best... we could have the 36_patch1 with slc6 and use in the 8_X MC Campaigns and have another one with UL (8_0_35_UL) that go under slc7... what do u think?

qliphy:
> @qliphy , see [cms-sw/cmssw#28655](https://github.com/cms-sw/cmssw/issues/28655) . May be we should build CMSSW_8_0_36_UL for slc7_amd64_gcc530?

@smuzaffar   I found 8_0_34 is identical as 8_0_34_UL,  except the former for slc6 while the latter for slc7.   However,for 8_0_35 and above, only non-UL release was built,  and the releases supports both slc6 and slc7.

That is why I think we should just build 8_0_36_patch1 instead of 8_0_36_UL  (we don't have 8_0_35_UL).  What do you and others think?

smuzaffar:
@qliphy  , looks like slc6 8_0_X is used for MC Campaigns so changing the production arch to slc7 might have a big impact there.

qliphy:
> hi guys, I would say that this is the best... we could have the 36_patch1 with slc6 and use in the 8_X MC Campaigns and have another one with UL (8_0_35_UL) that go under slc7... what do u think?

Ok,  seems good to have  8_0_36_patch1 with slc6 for 8_X MC, and 8_0_35_UL (or 8_0_36_UL?) with slc7 for UL.

qliphy:
> @qliphy , looks like slc6 8_0_X is used for MC Campaigns so changing the production arch to slc7 might have a big impact there.

Right.  I agree.  Then let's close this PR. 
And I will launch the build of  CMSSW_8_0_36_UL for slc7_amd64_gcc530 following  cms-sw/cmssw#28655 
BTW: @smuzaffar  should it be CMSSW_8_0_36_UL or CMSSW_8_0_35_UL?

smuzaffar:
clean solution would be that we build 
- CMSSW_8_0_36_UL for slc7 using CMSSW_8_0_36 tag ( just to be consistent with CMSSW_8_0_34_UL naming )
- CMSSW_8_0_36_UL_patch1 for slc7

what do you think?

jordan-martins:
good from my side! :)

qliphy:
Thanks @smuzaffar  Sounds good to me.  Let's do like this.  

Just to make sure, we will do two builds as following, right? :

1)   Build CMSSW_8_0_36_UL
CMSSW_COMMIT:CMSSW_8_0_36
ARCHITECTURE:slc7_amd64_gcc530
PRODUCTION_ARCHITECTURE:slc7_amd64_gcc530

2)  Build CMSSW_8_0_36_UL_patch1
ARCHITECTURE:slc7_amd64_gcc530
PRODUCTION_ARCHITECTURE:slc7_amd64_gcc530

smuzaffar:
@qliphy , yes this is correct. I am closing this PR now

PR #1692, opened 2022-01-19
Update categories.py

perrotta:
Remove agrohsje from "generators"


agrohsje:
Thanks @perrotta. Indeed that is what I had in mind.

PR #1696, opened 2022-01-23
ignore 0C402854-EECC-E311-AE34-E0CB4E553656.root file from workflow 4.76

smuzaffar:
This PR is work around to ignore `0C402854-EECC-E311-AE34-E0CB4E553656.root` file which is not available any more and causes workflow 4.76 to fail in all IBs. This change should allow to run workflow 4.76 in all IBs. Proper fix should be to fix workflow 4.76 configuration and use an available dataset


smuzaffar:
test parameters:
- workflow_threading = 4.76
- enable_test = threading

smuzaffar:
please test

smuzaffar:
ignore one file does not solve the issue, we need to come up with a proper solution

smuzaffar:
abort

PR #1706, opened 2022-02-03
Update phase2 profiling workflow to 35234.21

gartung:
D77 ttbar prodlike PU


smuzaffar:
test parameters:
- enable_test = profiling

smuzaffar:
please test

smuzaffar:
@gartung this is good to go in. By the way why to change the workflow, was there a request to use for the change?

gartung:
Yes. @perrotta suggested it.

smuzaffar:
+externals

perrotta:
> Yes. @perrotta suggested it.

Thank you Patrick, and Shahzad

To elaborate a bit further, we realized together with @cms-sw/reconstruction-l2  and @cms-sw/upgrade-l2 conveners that the D49 geometry used for profiling the phase2 workflows in the IBs and in the PR tests was not the currently preferred one for upgrade studies and developments (in particular by MTD, for which @fabiocos originally posed the question). Reco uses D76 for its performance tests, D77 is used in RelVals (and it will move soon to D88). D88 samples with PU will only become available after 12_3_0_pre5, therefore D77 looks now the best possible benchmark for performance measurements also for the IBs, before moving to D88.

PR #1709, opened 2022-02-04
Remove EventFilter/Utilities from reco

jpata:
In https://github.com/cms-sw/cms-bot/pull/485 reco was added as a signature to EventFilter/Utilities, but it is my understanding that we only follow one file in this package: https://github.com/cms-sw/cmssw/blob/master/EventFilter/Utilities/plugins/TcdsRawToDigi.cc.

Here I propose to remove the reco sig from the full package.
Is there a way to follow automatically only one file from the package?

@cms-sw/reconstruction-l2

### DIFF
--- a/categories_map.py
+++ b/categories_map.py
@@ -1135,7 +1135,6 @@ CMSSW_CATEGORIES = {
     "EventFilter/SiPixelRawToDigi",
     "EventFilter/SiStripRawToDigi",
     "EventFilter/TotemRawToDigi",
-    "EventFilter/Utilities",
     "JetMETCorrections/Configuration",
     "JetMETCorrections/JetCorrector",
     "JetMETCorrections/Modules",


smuzaffar:
@jpata , bot only keeps track of cmssw packages so there is no way to watch a single file.

jpata:
What about also moving `TcdsRawToDigi.cc` to `EventFilter/OnlineMetaDataRawToDigi` which both reco and DAQ are watching, and where the plugin is actually used?

smuzaffar:
sounds reasonable, `TcdsRawToDigi.cc` is plugin so should be trivial to move to `EventFilter/OnlineMetaDataRawToDigi` .  @cms-sw/daq-l2 any objections?

smorovic:
@smuzaffar,
no objections. It did seem out of place in `Utilities` and `OnlineMetaDataRawToDigi` is a better location.
Let me know if you want me to submit a PR or you will make the change.

smorovic:
I am preparing a PR for master / 12_3_X (let me know if in case you're trying to do the same).

smorovic:
Here is PR I just opened:
https://github.com/cms-sw/cmssw/pull/36887

smorovic:
@cmsbuild please test

smorovic:
testing wrong PR..

jpata:
https://github.com/cms-sw/cmssw/pull/36887 was merged, so possibly this can move forward?

PR #1712, opened 2022-02-17
update IB igprof runs to use a 2018 wf 136.889 instead of a 2016

slava77:
this is to follow up on cms-sw/cmssw#36282

### DIFF
--- a/schedule-additional-tests
+++ b/schedule-additional-tests
@@ -19,7 +19,7 @@ for x in `grep ADDITIONAL_TESTS=  $CONFIG_MAP | grep "SCRAM_ARCH=$ARCHITECTURE;"
     prof=$(echo $x | sed 's|igprof-||')
     sed -i -e "s|ADDITIONAL_TEST_NAME=$x|ADDITIONAL_TEST_NAME=igprof|" ${pfile}
     echo "PROFILE=$prof" >> ${pfile}
-    wf="136.731"
+    wf="136.889"
     nfile=$WORKSPACE/auto-$x-${wf}-properties
     cp ${pfile} ${nfile}
     echo "WORKFLOWS=-i all -l $wf" >> ${nfile}


slava77:
@smuzaffar 
I'm not sure how to test this one; I suppose that `please test` will not be consequential.

I guess it's something picked up in https://cmssdt.cern.ch/jenkins/job/ib-schedule-additional-tests/

gartung:
@slava77 Adding the workflow to this line should enable the Piechart creation
https://github.com/cms-sw/cms-bot/blob/16dface76c332c3f05f50575ed16ec53d0db437d/cmssw-pr-test-config#L3

slava77:
> @slava77 Adding the workflow to this line should enable the Piechart creation
> 
> https://github.com/cms-sw/cms-bot/blob/16dface76c332c3f05f50575ed16ec53d0db437d/cmssw-pr-test-config#L3

@gartung 
Thank you.
I added it in a separate PR.

Are the parameters in cmssw-pr-test-config going to affect the IB tests as well?

If so, to avoid a duplicate workflow running, then this PR #1712 will not be necessary.

gartung:
This PR add changes the workflow run by the IB Igprof jenkins jobs. These are a martrix of cpu and memory jobs for each workflow in the list.

gartung:
Changing the line I mentioned affects to Jenkins Profiling jobs which produces the Piechart.

smuzaffar:
the change here only effects the IB's `igprof` jobs. I have started `igprof-mp and igprof-pp` jobs for this new workflow.
- https://cmssdt.cern.ch/jenkins/job/ib-run-igprof/27874/
- https://cmssdt.cern.ch/jenkins/job/ib-run-igprof/27875/
Once these jobs run fine then we can merge this.

smuzaffar:
both jobs ran file, so this look good for IB igprog jobs.

PR #1714, opened 2022-02-17
reco comparisons: add maps for Run3 data workflows

slava77:
to catch up with the updates added with cms-sw/cmssw#36942 in runTheMatrix.py


slava77:
- the warnings about the missing workflows are gone, as expected.
- the logs in the corresponding workflows make sense (jobs complete and some plots are made, which indicate that the mapping to the input root file names and process names is correct)

smuzaffar:
+externals

PR #1713, opened 2022-02-17
add 136.889 to pr profiling  wf list

slava77:
this is in a follow up to https://github.com/cms-sw/cms-bot/pull/1712#issuecomment-1043295212

PR #1715, opened 2022-02-21
Fix for formating of PR profiling report.

gartung:


gartung:
enable profiling

gartung:
please test with #1713

smuzaffar:
+externals

PR #1721, opened 2022-03-08
switch reco release profiling to 35234.21

jpata:
- use the new 35234.21 Phase2 workflow for reco release profiling.
- remove the outdated workaround scripts to load a newer igprof
- only used by reco, so no effect expected on anything else


jpata:
yes to both. the new workflow ran succesfully at https://cmssdt.cern.ch/jenkins/job/release-run-reco-profiling/174/parameters/

PR #1727, opened 2022-03-10
Add ml-prod to groups.yaml.

riga:
Hi!

I hope I added this at the right spot.
Of course, feel free to rename in case there is a common pattern that I missed.

@yongbinfeng

### DIFF
--- a/groups.yaml
+++ b/groups.yaml
@@ -23,3 +23,6 @@ trk-dqm:
 tau-pog:
   - mbluj
   - azotz
+ml-prod:
+  - yongbinfeng
+  - riga


smuzaffar:
@riga , if this group is needed to watch changes (get github notifications) when a change is requested for some selected packages then yes this is right place but you also need to update either `category-watchers.yaml` (if you want to watch all packages changes belonging to a [category](https://github.com/cms-sw/cms-bot/blob/master/categories_map.py)) or `watchers.yaml` (if you want to watch only a selected packages).

riga:
I think we would continue to configure watched packages on a per-user basis, but I'll check that again and maybe create a second PR in the future.

Just to be sure, the entry in groups.yaml allows people to contact us by including `@ml-prod`, right?

smuzaffar:
> Just to be sure, the entry in groups.yaml allows people to contact us by including @ml-prod, right?

No, this is not true. Note that `ml-prod` could be a valid github user and by doing `@ml-prod` you basically notify that person :-)

riga:
Ok, I had that suspicion already ... then we'll get in touch with other groups on best practices for that. Thank you!

Feel free to close :-)

makortel:
How about `@cms-sw/ml-l2` team?

PR #1729, opened 2022-03-14
Add Hui Wang as HCAL DPG L2

mseidel42:
Adding @wang-hui as HCAL DPG L2 :-)


smuzaffar:
any objections @wang-hui

wang-hui:
> any objections @wang-hui ?

No objections. Thank you :)

PR #1734, opened 2022-03-23
reco comparisons: add a map for 39434.911 D88 DD4hep workflow

slava77:
a new workflow was added in CMSSW_12_4_X_2022-03-18-2300 with cms-sw/cmssw#37264

this PR adds a mapping for this new workflow and should resolve the warnings appearing in the reco comparisons summary


slava77:
@cmsbuild please test

PR #1735, opened 2022-03-24
drop usage of python3or2 and find the correct python to run

smuzaffar:
these script are those which we run during PR tests and IB build. We are missing `python3` on `slc6` and `python` on `rhel8 & cs9` . So initially we created a wrapper script `python3or2` and include it in `PATH` now we propose to replace it with this special code fragment to avoid dependency on external `python3or2`


smuzaffar:
please test

smuzaffar:
please test for CMSSW_9_4_X

PR #1754, opened 2022-04-26
reco comparisons: add a map for 39434.75_TTbar_14TeV+2026D88_HLT75e33

slava77:
to match the update in the short matrix list after cms-sw/cmssw#37324 was added in CMSSW_12_4_X_2022-04-20-1100


slava77:
@cmsbuild please test

smuzaffar:
+externals

PR #1757, opened 2022-05-02
Make Python scripts conform to PEP-8

iarspider:
Also, replace `optparser` with `argparse` (notice: requires python 2.7 or 3.2+). We should also check if files marked with "Is this file used" are used or not.

Notable changes:
* I couldn't find any references to `ActionError` in Python documentation
* `runCmd` (i.e. [`subprocess.getstatusoutput`](https://docs.python.org/3/library/subprocess.html) ) does not throw any exceptions, even if the command fails, so surrounding it in `try..catch` is useless

@smuzaffar I'd like to get this done first, this will help with the removal of ignominy. Question: do we want to maintain compatibility with Python2? If so, with which versions of Python 2?


smuzaffar:
> Question: do we want to maintain compatibility with Python2?

Yes, as long as we have CMSSW releases which use python2 as default, we have to maintain python2 compatibility. Though for now all the scripts but the scripts we run in cmssw env needs to be py2/py3 compatible.

> If so, with which versions of Python 2?
default python version in `slc6` containers  and above i.e.
```
[muzaffar@cmsdev30 muz]$ cmssw-slc6
Singularity> python -V
Python 2.6.6
```

smuzaffar:
> Question: do we want to maintain compatibility with Python2?

Yes, as long as we have CMSSW releases which use python2 as default, we have to maintain python2 compatibility. Though for now all the scripts but the scripts we run in cmssw env needs to be py2/py3 compatible.

> If so, with which versions of Python 2?
default python version in `slc6` containers  and above i.e.
```
[muzaffar@cmsdev30 muz]$ cmssw-slc6
Singularity> python -V
Python 2.6.6
```

iarspider:
> Python 2.6.6
Ouch.

checkLibDeps.py
```diff
@@ -43,7 +44,10 @@ def checkPkg(self, pkg):
 
 
 def main():
-    import argparse
+    try:
+        import argparse
+    except ImportError:
+        import argparse_py26 as argparse
```
smuzaffar, line 50: @iarspider , this does not work in `cmssw-slc6` container
```
cmssw-slc6 -- python -c 'import argparse_py26 as argparse'
  File "<string>", line 1, in <module>
ImportError: No module named argparse_py26
```

iarspider: @smuzaffar it should work if you run it in cms-bot area (the backported module is a part of this PR: https://github.com/cms-sw/cms-bot/pull/1757/files#diff-b2463dea70353cc2d26e61467bc8cf63af14e43550da15fe9d56ca2ad33cb646)

smuzaffar: I would suggest to drop `argparse_py26`  use the `archived_argparse` ( e.g see /cvmfs/cms.cern.ch/share/overrides/python/tweak_program_helpers.py ). `/cvmfs/cms.cern.ch/share/overrides/python` is in PYTHONPATH once we have cmsenv.

iarspider:
I am done for now. @smuzaffar is there an easy way to test this code - just in case I did something wrong?

smuzaffar:
@iarspider , most of these changes are used by IB test jobs. In order to test these you can create a new branch in cms-sw/cms-bot and then update cms-bot/config.map to add `CMS_BOT_BRANCH=your-branch;` for any IB e.g. `12_4_DEVEL_X`. You can also add this for 5.3.X IBs to test the changes on slc6

PR #1758, opened 2022-05-04
Add `db` signature to DQM/BeamMonitor

francescobrivio:
**PR Description**
This PR adds the `db` signature to the `DQM/BeamMonitor` package.
The rational behind this addition is that the BeamMonitor code is currently the only DQM client running live in P5 that has directl access to upload payloads to CondDB, hence we (@cms-sw/db-l2) would like to add the `db` signature in order to be informed of any possible change.

I hope our DQM colleagues (@cms-sw/dqm-l2) agree with this proposal.

**EDIT**
I profit of this PR to update my personal watchlist.

### DIFF
--- a/categories_map.py
+++ b/categories_map.py
@@ -470,6 +470,7 @@ CMSSW_CATEGORIES = {
     "CondTools/SiPixel",
     "CondTools/SiStrip",
     "CondTools/Utilities",
+    "DQM/BeamMonitor",
     "OnlineDB/CSCCondDB",
     "OnlineDB/EcalCondDB",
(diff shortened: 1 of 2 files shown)


tvami:
@francescobrivio I think those list are supposed to be alpha-ordered, i.e. the DQM package should go after all the C* packages, not after O*

francescobrivio:
> @francescobrivio I think those list are supposed to be alpha-ordered, i.e. the DQM package should go after all the C* packages, not after O*

ahahah yea you are right! Fixed now in 4eb7799!

PR #1760, opened 2022-05-05
Drop any references to ignominy

iarspider:
ignominy was using information from old, perl-based SCRAM, no longer works with python-based one.


smuzaffar:
@iarspider , i think you can also drop `buildSetTest.py`

PR #1777, opened 2022-06-16
copy ~/tnsnames.ora to workspace see INC3162820 SNOW

smuzaffar:


smuzaffar:
please test

smuzaffar:
+externals
looks good.
@aandvalenzuela , this should fix the issue of mounting afs files under singularity on lxplus. although it is issue with lxplus but with this trick we can avoid afs dependency

PR #1780, opened 2022-06-21
move slava77 from reco L2 to tracking-pog, keep the PR test role

slava77:
@smuzaffar 
do I also need to update the cmssw_l2/l2.json or is it automated?

### DIFF
--- a/categories.py
+++ b/categories.py
@@ -19,7 +19,7 @@ APPROVE_BUILD_RELEASE =  list(set([ "smuzaffar"] + CMSSW_L1 + SPECIAL_RELEASE_MA
 #List of gh users who can request to build a release.
 REQUEST_BUILD_RELEASE = APPROVE_BUILD_RELEASE
 #List og gh users who are allowed to trigger Pull Request testing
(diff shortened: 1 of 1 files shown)


smuzaffar:
> do I also need to update the cmssw_l2/l2.json or is it automated?

no that will be updated automatically

categories.py
```diff
@@ -131,7 +130,7 @@
   "kshcheli":         ["proton-pog"],
   "kandrosov":        ["tau-pog"],
   "alebihan":         ["tau-pog"],
-  "mtosi":            ["tracking-pog"],
```
smuzaffar, line 134: oops, did not notice this change. @slava77 was this an intentional update?

slava77: yes, it was intentional.
@mtosi @mmusich @vmariani 
may want to clarify if there were some longer term arrangements

mmusich: there have been no such arrangements.

smuzaffar: so all is good then

PR #1795, opened 2022-07-21
Add Manos Vourliotis in the tracking pog watchers

mmusich:
Since he'll be starting as new Upgrade Tracking L3 convener.
FYI @cms-sw/tracking-pog-l2


mmusich:
@VourMa FYI

PR #1800, opened 2022-07-28
Added DataFormats/TotemReco to reconstruction category

forthommel:
Replaced (unused) DataFormats/TotemDigi to DataFormats/TotemReco used in https://github.com/cms-sw/cmssw/pull/38886

### DIFF
--- a/categories_map.py
+++ b/categories_map.py
@@ -1116,6 +1116,7 @@ CMSSW_CATEGORIES = {
     "DataFormats/SiStripCommon",
     "DataFormats/TCDS",
     "DataFormats/TauReco",
+    "DataFormats/TotemDigi",
     "DataFormats/TotemReco",
     "DataFormats/TrackCandidate",
     "DataFormats/TrackReco",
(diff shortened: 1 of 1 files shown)


makortel:
I see `DataFormats/TotemDigi` was used within 8_1_X. IIUC old directories should be kept here, but this case is limited to one release cycle, and that release cycle has no further developments (we haven't built IBs for 8_1_X for ages). I would nevertheless suggest to only add a new directory at this point, and check if `DataFormats/TotemDigi` can be removed with @smuzaffar when he gets back.

forthommel:
@makortel Many thanks for the fast reaction!
Thanks also for the clarification. I am now updating this PR to add back the TotemDigi subpackage monitoring, and leave its extraction for a possible followup outdated packages cleaning.

aandvalenzuela:
Hi @forthommel, let's merge this one and I will bring up the removal of `DataFormats/TotemDigi` with Shahzad once he is back, as suggested.

smuzaffar:
@forthommel , @makortel has explained is very well. We do not delete cmssw package to category mapping if package is used by any cmssw releas/PR as then bot can start setting `new-package-pending` labels for PRs where this packages was updated.

PR #1810, opened 2022-08-11
move `MuonAnalysis/MomentumScaleCalibration` from analysis to alca, in absence of an analysis signature

mmusich:
As per comment https://github.com/cms-sw/cmssw/pull/39029#issuecomment-1211900500
and as it is close enough.
@cms-sw/alca-l2 please confirm if you agree.


tvami:
> https://github.com/orgs/cms-sw/teams/alca-l2 please confirm if you agree.

Agreed

PR #1817, opened 2022-09-01
[Jenkins parser job] Unittest for jobs configuration

aandvalenzuela:
This unit test does the following checks:
1. All jobs defined in the config file exists in Jenkins.
2. All error categories have been properly defined in the errorMsg section where the regex are defined.
3. The parser can recognize all the defined actions.

### DIFF
--- a/.github/workflows/parser-config.yaml
+++ b/.github/workflows/parser-config.yaml
@@ -9,16 +9,13 @@ on:
 jobs:
   configfile-sanity-checks:
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        python-version: [ '2.x', '3.x' ]
-    name: Python ${{ matrix.python-version }}
     steps:
-      - uses: actions/checkout@v2
-      - name: Set up Python
+      - name: Clone github repo
+        uses: actions/checkout@v2
+      - name: Setup python3
         uses: actions/setup-python@v2
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: '3.x'
           architecture: x64
       - name: Checking job's configuration in parser job
         run: PYTHONPATH=. python jenkins/parser/paser-config-unittest.py


jenkins/parser/paser-config-unittest.py
```diff
@@ -0,0 +1,50 @@
...
+print("[TEST 1]: Checking that all jobs defined in jobs-config.json exist in Jenkins ...")
+job_names = [jenkins_jobs[job_id]["jobName"] for job_id in range(len(jenkins_jobs))]
+
+for job in job_names:
+    exit_code = os.system(
+        "ssh -o IdentitiesOnly=yes -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i /var/lib/jenkins/.ssh/id_rsa-openstack -l localcli -p 8090 localhost get-job "
+        + job
+        + " > /dev/null"
+    )
+    assert exit_code == 0, "Job " + job + " does not exists in Jenkins"
```
smuzaffar, line 22: @aandvalenzuela , I do not think this check is going to work as github is not going to run this on jenkins server node.  I think you can download https://raw.githubusercontent.com/cms-sw/cmssdt-wiki/master/jenkins_reports/All.md and then search for job names in it

jenkins/parser/paser-config-unittest.py
```diff
@@ -1,5 +1,9 @@
 import json
 import os
+import re
+from subprocess import getstatusoutput
```
smuzaffar, line 4: @aandvalenzuela , this is good for python3 but will fail for python2. You can either do something like
```
import sys
if sys.version_info[0] == 2:
  from commands import getstatusoutput
else:
  from subprocess import getstatusoutput
```

or better to drop the python2 support and update the github yaml file to only run the test for python3

aandvalenzuela: Dropping python2 support, thanks!

PR #1819, opened 2022-09-02
Jenkins log parser: additional string for git failures

iarspider:
Detect `Connection to github.com closed by remote host.` message and treat it as git failure.

### DIFF
--- a/jenkins/parser/jobs-config.json
+++ b/jenkins/parser/jobs-config.json
@@ -100,7 +100,7 @@
         "errorStr": [
           "fatal: repository .* not found",
           "Tag already exist",
-          "Connection to github\.com closed by remote host."
+          "Connection to github\\.com closed by remote host."
         ],
         "action": "retryBuild"
       },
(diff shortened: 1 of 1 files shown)


smuzaffar:
@aandvalenzuela , can you please check why test failed here

aandvalenzuela:
@iarspider, you need to indicate also for which jobs you want to activate the check of `gitErrors`. This is done for each job where you want to activate the check in section `errorType`.

PS: I am planning to add a generic section to add the categories that must be checked for all jobs so that we don't need to modify every job if we want to add this error category in all of them.

aandvalenzuela:
@iarspider, I have added the category `gitErrors` to the jobs failing today for this reason in https://github.com/cms-sw/cms-bot/pull/1822

iarspider:
@aandvalenzuela thanks!

aandvalenzuela:
@iarspider, we also need to merge this one with the new string

PR #1825, opened 2022-09-05
updating xpog members

vlimant:


vlimant:
how cool would it be to simply synchronise the category content to the e-group listing @smuzaffar !

smuzaffar:
@vlimant , category content needs github user name while e-groups contain cern linux users or emails:-)

PR #1827, opened 2022-09-06
clean-up TRK DPG / POG categories

mmusich:
update TRK DPG and TRK POG lists with conveners as of 1st Sept 2022.
@vmariani @sroychow @tsusa please shout if you disagree ;)


vmariani:
Thanks @mmusich

PR #1835, opened 2022-09-08
Add syuvivida as new DQM L2 convener

francescobrivio:
Adding @syuvivida to `dqm` signature as new DQM L2 convener.

FYI @cms-sw/dqm-l2


emanueleusai:
+1

PR #1845, opened 2022-09-29
[Jenkins parser] Add options to force retry and to check concrete error message in all jobs

aandvalenzuela:
The fix is already running in Jenkins for testing purposes in https://cmssdt.cern.ch/jenkins/job/jenkins-test-parser/17230/ and runs successfully.

It fixes jobs like [ib-run-HLT](https://cmssdt.cern.ch/jenkins/job/ib-run-HLT/), where Jenkins is configured to self-retry if the build fails, but it doesn't schedule the retry if the build fails because of connection issues.

@smuzaffar, do you have any other suggestion rather than hardcoding the regex _Remote call on .* failed_?

### DIFF
--- a/jenkins/parser/jenkins-parser-job.py
+++ b/jenkins/parser/jenkins-parser-job.py
@@ -136,24 +136,25 @@ def get_running_builds(job_dir, last_processed_log):
     ]
 
 
-def trigger_retry_action(job_to_retry, build_to_retry, build_dir_path, action):
-    # Skip autoretry if Jenkins already retries
-    if grep(os.path.join(build_dir_path, "build.xml"), "<maxSchedule>", True):
-        print("... Jenkins already takes care of retrying. Skipping ...")
-        if grep(os.path.join(build_dir_path, "build.xml"), "<retryCount>", True):
+def trigger_retry_action(job_to_retry, build_to_retry, build_dir_path, action, regex):
+    # Skip autoretry if Jenkins already retries, unless connection issue.
+    if "Remote call on .* failed" not in regex:
+        if grep(os.path.join(build_dir_path, "build.xml"), "<maxSchedule>", True):
+            print("... Jenkins already takes care of retrying. Skipping ...")
+            if grep(os.path.join(build_dir_path, "build.xml"), "<retryCount>", True):
+                return
+            # Update description of the failed job
+            update_label = (
+                os.environ.get("JENKINS_CLI_CMD")
+                + " set-build-description "
+                + job_to_retry
+                + " "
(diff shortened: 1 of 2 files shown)


smuzaffar:
@aandvalenzuela , how about adding a option to the `errorMsg` e.g. `force: "true"`.By default it should be `false` but if an erorMsg sets it to `true` then we force retry that job?

aandvalenzuela:
As suggested above, there is a new filed in `jobs-config.json` (`forceRetry`) that can be set to true, so that the parser retries the build although Jenkins is configured to retry it as well. It fixes jobs like [ib-run-HLT](https://cmssdt.cern.ch/jenkins/job/ib-run-HLT/), where Jenkins is configured to self-retry if the build fails, but it doesn't schedule the retry if the build fails because of connection issues.

In addition, the keyword `allJobs` can be also set to true so that the corresponding error category applies to all the jobs monitored by the parser.

jenkins/parser/jenkins-parser-job.py
```diff
@@ -23,10 +23,22 @@ def grep(filename, pattern, verbose=False):
...
+    for ii in jobs_object["jobsConfig"]["errorMsg"].keys():
+        if jobs_object["jobsConfig"]["errorMsg"][ii]["allJobs"] == "true":
+            common_keys.append(ii)
+        if jobs_object["jobsConfig"]["errorMsg"][ii]["forceRetry"] == "true":
+            force_retry_regex.extend(
+                jobs_object["jobsConfig"]["errorMsg"][ii]["errorStr"]
+            )
+    # Get the error keys of the concrete job ii
     error_keys = jobs_object["jobsConfig"]["jenkinsJobs"][job_id]["errorType"]
+    error_keys.extend(common_keys)
```
smuzaffar, line 40: @aandvalenzuela , shouldn't we make a copy of `jobs_object["jobsConfig"]["jenkinsJobs"][job_id]["errorType"]` in above line? e.g. 
```
error_keys = jobs_object["jobsConfig"]["jenkinsJobs"][job_id]["errorType"][:]
```
I think otherwise `error_keys.extend(common_keys)` is going to add the new commonkeys in the original `job_object`

smuzaffar:
@aandvalenzuela  looks good to me. Have you tested these changes? Can we merge this or should we do it on Monday?

aandvalenzuela:
They are currently running in https://cmssdt.cern.ch/jenkins/job/jenkins-test-parser/17244/ with extra prints and looks good.

We can merge, I will keep an eye on in case we hit a corner case.

PR #1857, opened 2022-10-13
Disabled 12.2.X and 12.3.X IBs

smuzaffar:
As suggested during ORP meeting https://indico.cern.ch/event/1208162/ , this change disables the build of 12.2.X and 12.3.X IBs.


smuzaffar:
merging it as there is nothing to test here

PR #1860, opened 2022-10-13
reco comparisons new phase-2 wf numbering after reset around 2022-10-11

slava77:
- index reset was introduced in cms-sw/cmssw#39571 
- D94 workflows were added in cms-sw/cmssw#38702 , but I did not cover the old numbering variant

this should clear the warnings in the comparisons summaries


slava77:
@cmsbuild please test

slava77:
@cms-sw/upgrade-l2 @cms-sw/reconstruction-l2

slava77:
> Comparison Summary

it looks like all 6 workflows for phase2 were recognized and comparisons directories with logs show up as expected.

smuzaffar:
+externals

Issue #1854, opened 2022-10-13
Workflow `39634.21` undefined in `igprof` 12_6_X


aandvalenzuela:
Hello,

We are having some failures in `igprof` jobs due to workflow `39634.21` not defined: 
```
05:29:28   File "/tmp/cmsbuild/jenkins/workspace/ib-run-igprof/CMSSW_12_6_X_2022-10-12-2300/bin/el8_amd64_gcc10/runTheMatrix.py", line 606, in <module>
05:29:28     ret = runSelected(opt)
05:29:28   File "/tmp/cmsbuild/jenkins/workspace/ib-run-igprof/CMSSW_12_6_X_2022-10-12-2300/bin/el8_amd64_gcc10/runTheMatrix.py", line 31, in runSelected
05:29:28     if len(undefSet)>0: raise ValueError('Undefined workflows: '+', '.join(map(str,list(undefSet))))
05:29:28 ValueError: Undefined workflows: 39634.21
05:29:28 ++ ls -d '39634.21*'
05:29:28 ls: cannot access '39634.21*': No such file or directory
05:29:28 + mv runTheMatrix.log
05:29:28 mv: missing destination file operand after 'runTheMatrix.log'
05:29:28 Try 'mv --help' for more information.
05:29:28 + ERR=1
05:29:28 + false
05:29:28 + exit 1
05:29:28 + ERR=1
```
See complete stack trace: https://cmssdt.cern.ch/jenkins/job/ib-run-igprof/37997/console
Failures appear only in in `CMSSW_12_6_X`. Could this be related with the clean-up of workflows performed in `12_6_X`?

I also see that this workflow has been hardcoded in [cms-cmpwg/profiling](https://github.com/cms-cmpwg/profiling/blob/master/Gen_tool/Gen.sh#L40). @gartung I see you are regularly taking care of this repository :) 

Could you please have a look?

Thank you,
Andrea.

makortel:
> Could this be related with the clean-up of workflows performed in `12_6_X`?

I think this workflow is one of the phase2 workflows that were renumbered recently.  @srimanob could you confirm?

srimanob:
39634.21 is prod workflow of PU ttbar with D88. The workflow id is reset. The new id for this is 21034.21.

gartung:
The workflow for the Jenkins job is defined here
[...]

PR #1861, opened 2022-10-13
run-ib-igprof script: check if workflow exists in release 

gartung:
This will check if a workflow is defined in an IB release and warn if it is not, skipping the runTheMatrix.py command.

fixes https://github.com/cms-sw/cms-bot/issues/1854


gartung:
@aandvalenzuela there is no way to test this until the next IB is run but I checked it from the command line,

gartung:
please test

aandvalenzuela:
Thanks @gartung! :)

PR #1866, opened 2022-10-17
reco comparisons: map for renano wf 2500.601

slava77:
needed after cms-sw/cmssw#39472   which was merged in CMSSW_12_6_X_2022-10-14-1100

@cms-sw/reconstruction-l2 @cms-sw/xpog-l2


slava77:
@cmsbuild please test

comparisons/matrix_RE.txt
```diff
@@ -377,6 +377,7 @@ TTbarFS14TeVPU2021wf13434p0 13434.0_TTbar_14TeV+*2021*FS*PU*/step1.root RECO
 TTbar14TeV2023wf12434p0 12434.0_TTbar_14TeV+*2023*/step3.root RECO
 TTbar14TeVPU2023wf12634p0 12634.0_TTbar_14TeV+*2023PU*/step3.root RECO
 TTbar14TeVPU2023nanoEDMwf12634p0 12634.0_TTbar_14TeV+*2023PU*/step5.root NANO
+TTbar14TeVPU2023nanoEDMin126Xwf2500p601 2500.601_mc126X+TTBarMINIAOD*/step2.root NANO
```
smuzaffar, line 380: @slava77 , for my education, `TTbar14TeVPU2023nanoEDMin126Xwf2500p601` is just an arbitrary string which should be unique for ecah entry in this file ... right?

slava77: it's supposed to be a shorter representation of the full directory name with occasionally an insert/suffix related to the file/process/step in the workflows.
Uhm, 2023 is wrong though, it's a 2022 wf, let me fix that.

smuzaffar:
please test

smuzaffar:
+externals

vlimant:
@slava77 when adding 2500.601, I am suspecting that the scripts expects an edm file with `nanoaodFlatTable` branches, while the workflows are producing nano output directly.
Should I update `validate.C` to plot things from flat root files ?

slava77:
> @slava77 when adding 2500.601, I am suspecting that the scripts expects an edm file with `nanoaodFlatTable` branches, while the workflows are producing nano output directly. Should I update `validate.C` to plot things from flat root files ?

oh, I didn't realize that part. 

I have added the NANO monitoring to validate.C since it was almost free for maintenance: the number and names of FlatTables rarely changes and the list of plots is derived from the table content automatically.

Would the part that you are planning to add be similarly (if not better) automated? In that case it would be a good addition for `validate.C`. If it will require updates for every new branch, I'm less sure.

vlimant:
the update would auto discover all branches. I suspect I can also update the edm nano part to also discover all flat table products

PR #1870, opened 2022-10-24
[Parser] Add cleanup for cmssdt parser monitor page

aandvalenzuela:
This PR adds a cleanup method for the monitor page https://cmssdt.cern.ch/SDT/public/test-parser-web-info.html.
It includes some debug prints that I will remove in a follow-up PR.


aandvalenzuela:
Merging since it is already running in Jenkins.

PR #1862, opened 2022-10-14
Improve PR comparison job to run tests in parallel

smuzaffar:
Various improvements:
- `logRootQA.py` process logs and root files for IB and PR tests. This allow it to process these files while `JR` and `default-comparsion` tests are still running.
- properly handle the back ground jobs
- Compile root macros once

PR #1872, opened 2022-11-02
automatic finding of edm and pure root nano branches

vlimant:
as commented in https://github.com/cms-sw/cms-bot/pull/1866#issuecomment-1298570409 the current script is not parsing the edm flat table automatically, and is not compatible with pure-root nano output.
this PR tries to address both issues.
I might include the matrix_RE change for the new nano workflows

### DIFF
--- a/comparisons/validate.C
+++ b/comparisons/validate.C
@@ -3,6 +3,7 @@
 #include "TChain.h"
 #include "TCanvas.h"
 #include "TString.h"
+#include "TRegexp.h"
 #include "TFile.h"
 #include "TROOT.h"
 #include "TH1F.h"
@@ -58,6 +59,19 @@ bool stepContainsNU(const TString& s, TString v){
   } else {
     return s.Contains(v);
   }
+}
+
+std::vector<TString>  getBranchNames(const TString& bPattern){
+  TObjArray * all_branches = Events->GetListOfBranches();
+  TRegexp exp(bPattern);
+  std::vector<TString> branches;
+  branches.reserve(all_branches->GetSize());
+  for (int i=0 ; i!=all_branches->GetSize() ; ++i){
+    TString branch_name(all_branches->At(i)->GetName());
+    if (branch_name(exp).Length())
+      branches.push_back(branch_name);
+  }
+  return branches;
 }
 
 bool checkBranchAND(const TString& b, bool verboseFalse = false){
@@ -1475,7 +1489,10 @@ void evtPlanes(const TString& bName, bool detailByPlane = false){
   }
 }
 
-void flatTable(const TString& shortName){
+void flatTable(const TString& branchName){
+  plotvar(branchName, "", true);
+}
+void flatEDMTable(const TString& shortName){
   const TString bObj = "nanoaodFlatTable_"+shortName+"_"+recoS+".obj";
   if (! checkBranchOR(bObj, true)) return;
 
@@ -1641,9 +1658,33 @@ void validateEvents(TString step, TString file, TString refFile, TString r="RECO
 
   std::cout<<"Start making plots for Events with "<<Nnew<<" events and refEvents with "<<Nref<<" events ==> check  "<<Nmax<<std::endl;
 
+  if (stepContainsNU(step, "NANO")){
+    std::vector<TString> edmBranches = getBranchNames("nanoaodFlatTable_.*__"+recoS);
+    //Check if it's a NANOEDM
+    if (edmBranches.size()){
+      for (uint iT = 0; iT < edmBranches.size() ; ++iT){
+	TString shortName = ((*edmBranches[iT].Tokenize("_"))[1])->GetName();
+	shortName += "_";
+	//shortName += ((*edmBranches[iT].Tokenize("_"))[2])->GetName();
+	std::cout<< "this is a shortname"<< shortName <<std::endl;
+        flatEDMTable( shortName);
+      }
+      std::cout<<"Done comparing nano-EDM"<<std::endl;
+    }else{
+      //NANO was asked for, but no EDM branches. plot nano branches
+      std::vector<TString> nanoBranches = getBranchNames(".*");
+      for (uint iT = 0; iT <nanoBranches.size() ; ++iT){
+	flatTable( nanoBranches[iT] );
+      }
+      std::cout<<"Done comparing pure-ROOT nano"<<std::endl;
+    }
+
+    return;
+  }
+
   if (!stepContainsNU(step, "hlt")){
 
-    //Check if it's a NANOEDM
+    /*
     if (checkBranchAND("nanoaodFlatTable_muonTable__"+recoS+".")){
       const int nTabs = 68;
       TString tNames[nTabs] = {
@@ -1721,12 +1762,12 @@ void validateEvents(TString step, TString file, TString refFile, TString r="RECO
         "genProtonTable_",                       //67
       };
       for (int iT = 0; iT < nTabs; ++iT){
-        flatTable(tNames[iT]);
+        flatEDMTable(tNames[iT]);
       }
 
       std::cout<<"Done comparing nano-EDM"<<std::endl;
       return;
-    }
+      }*/
 
 
     if ((stepContainsNU(step, "all") || stepContainsNU(step, "error"))){


comparisons/validate.C
```diff
@@ -1641,9 +1658,33 @@ void validateEvents(TString step, TString file, TString refFile, TString r="RECO
...
+      std::cout<<"Done comparing pure-ROOT nano"<<std::endl;
+    }
+
+    return;
+  }
+
   if (!stepContainsNU(step, "hlt")){
 
-    //Check if it's a NANOEDM
+    /*
```
vlimant, line 1687: I'll remove the commented lines after feedback

comparisons/validate.C
```diff
@@ -60,6 +61,19 @@ bool stepContainsNU(const TString& s, TString v){
   }
 }
 
+std::vector<TString>  getBranchNames(const TString& bPattern){
+  TObjArray * all_branches = Events->GetListOfBranches();
```
vlimant, line 65: @slava77 I can augment the list of branches with the one matching in the refEvents

comparisons/matrix_RE.txt
```diff
@@ -371,7 +371,21 @@ TTbar14TeVPUppmx2021wf11834p99 11834.99_TTbar_14TeV+*2021PU*/step4.root RECO
 TTbar14TeVPUppmx2021nanoEDMwf11834p99 11834.99_TTbar_14TeV+*2021PU*/step6.root NANO
 TTbarFS14TeV2021wf13234p0 13234.0_TTbar_14TeV+*2021*FS*/step1.root RECO
 TTbarFS14TeVPU2021wf13434p0 13434.0_TTbar_14TeV+*2021*FS*PU*/step1.root RECO
-TTbar14TeVPU2022nanoEDMin126Xwf2500p601 2500.601_mc126X+TTBarMINIAOD*/step2.root NANO
+TTbar14TeVmc106Xul16v2nanowf2500p310 2500.31_mc106Xul17v2+TTbarMINIAOD*/step2.root NANO
+TTbar14TeVmc106Xul17v2nanowf2500p311 2500.311_mc106Xul17v2+TTBarMINIAOD*/step2.root NANO
+TTbar14TeVmc106Xul18v2nanowf2500p312 2500.312_mc106Xul17v2+TTbarMINIAOD*/step2.root NANO
+MuonEGdata106Xul16nanowf2500p330 2500.33_data106Xul16v2+MuonEG2016MINIAOD*/step2.root NANO
+MuonEGdata106Xul17nanowf2500p331 2500.331_data106Xul17v2+MuonEG2017MINIAOD*/step2.root NANO
+MuonEGdata106Xul18nanowf2500p332 2500.332_data106Xul18v2+MuonEG2018MINIAOD*/step2.root NANO
```
slava77, line 379: better move these up where run2 workflows are used

comparisons/matrix_RE.txt
```diff
@@ -371,7 +371,21 @@ TTbar14TeVPUppmx2021wf11834p99 11834.99_TTbar_14TeV+*2021PU*/step4.root RECO
...
+TTbar14TeVmc122XnanoV10wf2500p400 2500.4_mc122Xrun3_v10+TTbarMINIAOD*/step2.root NANO
+TTbar14TeVmc122Xnanowf2500p401 2500.401_mc122Xrun3+TTbarMINIAOD*/step2.root NANO
+TTbar14TeVmc124XnanoV10wf2500p500 2500.5_mc124Xrun3_v10+TTbarMINIAOD*/step2.root NANO
+TTbar14TeVmc124Xnanowf2500p501 2500.501_mc124Xrun3+TTbarMINIAOD*/step2.root NANO
+MuonEGdata124XnanoV10wf2500p510 2500.51_data124Xrun3_v10+MuonEG2022MINIAOD*/step2.root NANO
+MuonEGdata124Xnanowf2500p511 2500.511_data124Xrun3+MuonEG2022MINIAOD*/step2.root NANO
+TTbar14TeVmc126XnanoV10wf2500p600 2500.6_mc126X_v10+TTBarMINIAOD*/step2.root NANO
+TTbar14TeVmc126Xnanowf2500p601 2500p601 2500.601_mc126X+TTBarMINIAOD*/step2.root NANO
```
slava77, line 387:
```suggestion
TTbar14TeVmc126Xnanowf2500p601 2500.601_mc126X+TTBarMINIAOD*/step2.root NANO
```
looks like a typo

comparisons/validate.C
slava77: IIRC, ref is loaded and plotted first. So, using `refEvents` would be better.

smuzaffar:
enable nano

smuzaffar:
please test

comparisons/validate.C
```diff
@@ -1641,93 +1658,32 @@ void validateEvents(TString step, TString file, TString refFile, TString r="RECO
 
   std::cout<<"Start making plots for Events with "<<Nnew<<" events and refEvents with "<<Nref<<" events ==> check  "<<Nmax<<std::endl;
 
-  if (!stepContainsNU(step, "hlt")){
-
+  if (stepContainsNU(step, "NANO")){
```
slava77, line 1661: this part of the change would not work.
The driver shell script assigns everything to step `all` and it's up to `validateEvents` to figure out what to plot (in most cases just try everything).

It looks like the EDM files have some "standard" branches like `EventAuxiliary` or `EventProductProvenance` in the `Events` tree. Perhaps these could be used to guess the file content.

A more controllable alternative is to extend the input map file format to have one more field for the type of a file (nano, reco, or even mini). 
The current format of the map file `shortName fullName.root processName comments`.

vlimant: I thought, without checking, that the third was the `step` that was passed to the script indeed.
I’ll switch back to the initial if statement

vlimant:
I left out some of the workflows based on #39796

smuzaffar:
please test with cms-sw/cmssw#39796

smuzaffar:
please test for CMSSW_12_5_X

smuzaffar:
please test for CMSSW_12_4_X

smuzaffar:
please test for CMSSW_9_4_X

smuzaffar:
@vlimant @slava77 do you have any other changes planned for this PR? Thsi looks good to me

vlimant:
I cannot really tell whether the comparison was done properly ; I let @slava77 give the green light or else

comparisons/matrix_RE.txt
```diff
@@ -214,8 +221,13 @@ RunJetHT2018DecalGPUwf136p888512 136.888512_RunJetHT2018D+*ECALOnlyGPU*/step3.ro
 RunJetHT2018DhcalCPUwf136p888521 136.888521_RunJetHT2018D+*HCALOnlyCPU*/step3.root reRECO
 RunJetHT2018DhcalGPUwf136p888522 136.888522_RunJetHT2018D+*HCALOnlyGPU*/step3.root reRECO
 RunParkingBPH2018Bwf136p898 136.898_RunParkingBPH2018B+*/step3.root reRECO
+MuonEGdata106Xul18nanowf2500p332 2500.332_data106Xul18v2+MuonEG2018MINIAOD*/step2.root NANO
+
 #
 # Run3 data
+MuonEGdata124XnanoV10wf2500p510 2500.51_data124Xrun3_v10+MuonEG2022MINIAOD*/step2.root NANO
+MuonEGdata124Xnanowf2500p511 2500.511_data124Xrun3+MuonEG2022MINIAOD*/step2.root NANO
```
slava77, line 229:
```suggestion
MuonEG2022D124XnanoV10wf2500p51 2500.51_data124Xrun3_v10+MuonEG2022MINIAOD*/step2.root NANO
MuonEG2022D124Xnanowf2500p511 2500.511_data124Xrun3+MuonEG2022MINIAOD*/step2.root NANO
```
Please keep the origin information in the short names. Simply 124X is not enough.

comparisons/matrix_RE.txt
```diff
@@ -214,8 +221,13 @@ RunJetHT2018DecalGPUwf136p888512 136.888512_RunJetHT2018D+*ECALOnlyGPU*/step3.ro
 RunJetHT2018DhcalCPUwf136p888521 136.888521_RunJetHT2018D+*HCALOnlyCPU*/step3.root reRECO
 RunJetHT2018DhcalGPUwf136p888522 136.888522_RunJetHT2018D+*HCALOnlyGPU*/step3.root reRECO
 RunParkingBPH2018Bwf136p898 136.898_RunParkingBPH2018B+*/step3.root reRECO
+MuonEGdata106Xul18nanowf2500p332 2500.332_data106Xul18v2+MuonEG2018MINIAOD*/step2.root NANO
```
slava77, line 224:
```suggestion
MuonEG2018D106Xul18nanowf2500p332 2500.332_data106Xul18v2+MuonEG2018MINIAOD*/step2.root NANO
```

comparisons/matrix_RE.txt
```diff
@@ -180,6 +185,8 @@ RunJetHT2017Fwf136p831 136.831_RunJetHT2017F+*/step3.root reRECO
 RunJetHT2017FreMINIAODwf136p8311 136.8311_RunJetHT2017F_reminiaod+*/step2.root PAT
 RunJetHT2017FreMINIAODULwf136p83111 136.83111_RunJetHT2017F_reminiaodUL+*/step2.root PAT
 RunMET2017Fwf136p832 136.832_RunMET2017F+*/step3.root reRECO
+MuonEGdata106Xul17nanowf2500p331 2500.331_data106Xul17v2+MuonEG2017MINIAOD*/step2.root NANO
```
slava77, line 188:
```suggestion
MuonEG2017F106Xul17nanowf2500p331 2500.331_data106Xul17v2+MuonEG2017MINIAOD*/step2.root NANO
```

comparisons/matrix_RE.txt
```diff
@@ -159,6 +162,8 @@ RunJetHT2016Hwf136p772 136.772_RunJetHT2016H+*/step3.root reRECO
 RunJetHT2016HreMINIAODwf136p7721 136.7721_RunJetHT2016H_reminiaod+*/step2.root PAT
 RunJetHT2016HreMINIAODULwf136p77211 136.77211_RunJetHT2016H_reminiaodUL+*/step2.root PAT
 RunJetHT2016HnanoEDMwf136p7722 136.7722_RunJetHT2016H*NANOEDM*/step2.root DQM
+MuonEGdata106Xul16nanowf2500p330 2500.33_data106Xul16v2+MuonEG2016MINIAOD*/step2.root NANO
```
slava77, line 165:
```suggestion
MuonEG2016E106Xul16nanowf2500p33 2500.33_data106Xul16v2+MuonEG2016MINIAOD*/step2.root NANO
```

comparisons/matrix_RE.txt
```diff
@@ -125,6 +125,9 @@ TTbar13nanoEDMin2017wf1325p8 1325.8_TTbar_13_94X*NanoAOD*NANOEDM*/step2.root DQM
 TTbar13nanoEDMin2017wf1325p9 1325.9_TTbar_13_92X*NanoAOD*NANOEDM*/step2.root DQM
 TTbar13nanoEDM106Xv1in2017wf1325p81 1325.81_TTbar_13_106Xv1NanoAOD*NANOEDMMC2017*/step2.root DQM
 ZEE13nanoEDMin2016wf1329p1 1329.1_ZEE_13_80X*NanoAOD*NANOEDM*/step2.root DQM
+TTbarmc106Xul16v2nanowf2500p310 2500.31_mc106Xul17v2+TTbarMINIAOD*/step2.root NANO
+TTbarmc106Xul17v2nanowf2500p311 2500.311_mc106Xul17v2+TTBarMINIAOD*/step2.root NANO
+TTbarmc106Xul18v2nanowf2500p312 2500.312_mc106Xul17v2+TTbarMINIAOD*/step2.root NANO
```
slava77, line 130:
```suggestion
TTbar13PU106Xul16v2nanowf2500p31 2500.31_mc106Xul17v2+TTbarMINIAOD*/step2.root NANO
TTbar13PU106Xul17v2nanowf2500p311 2500.311_mc106Xul17v2+TTBarMINIAOD*/step2.root NANO
TTbar13PU106Xul18v2nanowf2500p312 2500.312_mc106Xul17v2+TTbarMINIAOD*/step2.root NANO
```
`mc` was not used before in the short names

Should the wf name be fixed at some point? `mc106Xul17v2` is in all 3 directory names while it should be 16,17,18, respectively. Thanks for making the map correct though

comparisons/matrix_RE.txt
```diff
@@ -371,7 +383,13 @@ TTbar14TeVPUppmx2021wf11834p99 11834.99_TTbar_14TeV+*2021PU*/step4.root RECO
 TTbar14TeVPUppmx2021nanoEDMwf11834p99 11834.99_TTbar_14TeV+*2021PU*/step6.root NANO
 TTbarFS14TeV2021wf13234p0 13234.0_TTbar_14TeV+*2021*FS*/step1.root RECO
 TTbarFS14TeVPU2021wf13434p0 13434.0_TTbar_14TeV+*2021*FS*PU*/step1.root RECO
-TTbar14TeVPU2022nanoEDMin126Xwf2500p601 2500.601_mc126X+TTBarMINIAOD*/step2.root NANO
+TTbar14TeVmc122XnanoV10wf2500p400 2500.4_mc122Xrun3_v10+TTbarMINIAOD*/step2.root NANO
+TTbar14TeVmc122Xnanowf2500p401 2500.401_mc122Xrun3+TTbarMINIAOD*/step2.root NANO
+TTbar14TeVmc124XnanoV10wf2500p500 2500.5_mc124Xrun3_v10+TTbarMINIAOD*/step2.root NANO
+TTbar14TeVmc124Xnanowf2500p501 2500.501_mc124Xrun3+TTbarMINIAOD*/step2.root NANO
+TTbar14TeVmc126XnanoV10wf2500p600 2500.6_mc126X_v10+TTBarMINIAOD*/step2.root NANO
+TTbar14TeVmc126Xnanowf2500p601 2500.601_mc126X+TTBarMINIAOD*/step2.root NANO
```
slava77, line 391:
```suggestion
TTbar13p6TeVPU2022in122XnanoV10wf2500p4 2500.4_mc122Xrun3_v10+TTbarMINIAOD*/step2.root NANO
TTbar13p6TeVPU2022in122Xnanowf2500p401 2500.401_mc122Xrun3+TTbarMINIAOD*/step2.root NANO
MinBias14TeV2022in124XnanoV10wf2500p5 2500.5_mc124Xrun3_v10+TTbarMINIAOD*/step2.root NANO
MinBias14TeV2022in124Xnanowf2500p501 2500.501_mc124Xrun3+TTbarMINIAOD*/step2.root NANO
TTbar14TeVPU2022in126XnanoV10wf2500p6 2500.6_mc126X_v10+TTBarMINIAOD*/step2.root NANO
TTbar14TeVPU2022in126Xnanowf2500p601 2500.601_mc126X+TTBarMINIAOD*/step2.root NANO
```
please keep the year or at least the era in the names.
Note that .5 and .501 use `/MinBias_TuneCP5_14TeV-pythia8/Run3Summer22MiniAODv3-NoPU_Pilot` as input. I'm not sure which was the intention. If the fix to the workflow is imminent, then leaving TTbar may be better, otherwise MinBias would be more clear when inspecting the differences

slava77 (commented):
I have some notes about the new workflow names, they are not really following the older patterns and frequently omit the period of data taking/simulation

comparisons/matrix_RE.txt
vlimant: with only one version of mini supported in input to nano in 124X, I think there is no need for a longer useless name.
please suggest a replacement though

comparisons/matrix_RE.txt
vlimant: there's a typo in the matrix wf name indeed. too cumbersome to correct ATM.
this mapping is evermore cumbersome IMO. a better way should be found to ease maintenance

comparisons/matrix_RE.txt
vlimant: it is intended to be a TTBar sample, indeed

vlimant:
the command 
`/afs/cern.ch/user/v/vlimant/work/cms-bot/comparisons/validateJR.py --base /data/cmsbld/jenkins/workspace/compare-root-files-short-matrix/data/PR-6c4b56 --ref /data/cmsbld/jenkins/workspace/compare-root-files-short-matrix/data/CMSSW_12_4_X_2022-11-07-1100 >> /data/cmsbld/jenkins/workspace/compare-root-files-short-matrix/results/JR-comparison/validateJR.log 2>&1`
would do

slava77:
can we split the nano branch plotting from the rest (that could've been an update only in validate.C file)? 

There are some downstream dependencies ( DQM comparisons running depends on parsing the fwlite/JR outputs)
Some workflows can also have multiple matching files and some processes are in general not a good match (e.g. `HLT` is present only in prehistoric fastim workflows).
Please make sure that the updates still work in at least 10_6_X

vlimant:
please test

slava77:
comparisons failed with 
```
10:59:14 cp: cannot stat '/data/cmsbld/jenkins/workspace/compare-root-files-short-matrix/cms-bot/comparisons/matrix_RE.txt': No such file or directory
```
need an update in  `pr_testing/run-pr-comparisons`

slava77:
> what do you mean by splitting form the rest? 

a separate PR. The removal of the map is a good effort, but is less trivial. It seemed like having a running comparison on NANO was a bit more timely/useful.

> what are the downstream dependencies ?

the now failed test is one example. There are more scripts in cms-bot that depend on the format of the JR outputs

> is that a good or bad thing to have multiple matching files ? intended or not ?

It takes time to try comparisons on a file.
validate.C doesn't break early, even though some "reliably present" branches could be checked.

Also, with the strict mapping you'd offhand know that every folder can have useful comparisons. If I parsed the script correctly, other step folders will appear as well with almost assuredly no useful output. Navigation may be harder.

Perhaps skip `step*inDQM.root` files already.

> I fail to understand how using the last process of an output file is not the proper way to look at the produced products of the last process ...

good point, I didn't read the script carefully enough


> I think this will also work in 10.6 can this be tested here ?

slava77:
> I think this will also work in 10.6 can this be tested here ?

this was more about python2 vs python3

vlimant:
all folders and output are parsed indeed (except for those containing 'inDQM','DQM_V') and only the one with the last process being one of 'ZStoRECO','RECO','reRECO','PAT','NANO','DQM','HLT','HLT2' will trigger a comparison.

vlimant:
please test

slava77:
It looks like this time it failed somewhere in comparisons/analyzeFWComparison.py

The console is not too clear on the details because the errors went to a log that got deleted
```
Started with default comparison at Fri Nov 11 13:51:02 CET 2022
+ ERR=1
+ false
+ exit 1
Build step 'Execute shell' marked build as failure
New run name is 'CMSSW_12_6_X_2022-11-10-2300 + cms-sw/cms-bot#1872 '
[WS-CLEANUP] Deleting project workspace...
[WS-CLEANUP] Skipped based on build state FAILURE
Finished: FAILURE

```
@smuzaffar is there a way to keep the outputs?

slava77:
> @smuzaffar is there a way to keep the outputs?

kind ping.
It's hard to debug the problems if a failing bot job does not save the logs.

aandvalenzuela:
Hi @slava77,
Sorry for the delay, Shahzad has been unavailable these last days. In principle, the tests results for this PR are stored at https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-6c4b56/ (where each folder corresponds to a different test run). The relevant results should be copied to this server before deleting the jenkins workspace. Nevertheless, I can also see that there are no results for the comparison itself. I have started a new run and I will keep all the logs and forward them to you.

Maybe there is a more direct way to get the logs, but in the mean time I think this should work.
I will double check that no important logs are deleted for this job in a regular basis.

Apologies for the inconveniences! Also note that you can ping all the team by using @cms-sw/externals-l2 :)

smuzaffar:
please test

smuzaffar:
@vlimant , does the new `validateJR.py` run jobs in parallel?

vlimant:
hi @smuzaffar, it does not, but could be made to

smuzaffar:
> hi @smuzaffar, it does not, but could be made to

the bash script was running the comparison in parallel so we should make sure that python one does the same. By the way, was there any thing which we could not have done in bash?

vlimant:
I don't speak bash fluently enough ;-)
I'll take a stab at parallelising

smuzaffar:
JR comparison now taking over 2h10 mins [a] while bash script was taking <15mins. I would suggest to keep the bash script and update it if needed. The bash script allowed us to run JR and other comparison jobs in parallel

[a] https://cmssdt.cern.ch/jenkins/job/compare-root-files-short-matrix/54099/consoleFull
```
Start JR comparison at Tue Nov 15 19:28:17 CET 2022
Finished with JR comparison at Tue Nov 15 21:41:01 CET 2022:
```

[b] https://cmssdt.cern.ch/jenkins/job/compare-root-files-short-matrix/54113/consoleFull
```
Start JR comparison at Wed Nov 16 07:06:52 CET 2022
Finished with JR comparison at Wed Nov 16 07:19:24 CET 2022:
```

vlimant:
@smuzaffar the `--procs N` is added, default would default to number of cores of the machine

vlimant:
please test

smuzaffar:
@vlimant , as I wrote earlier, with bash script we were able to run other jobs in parallel with JR comparisons. e.g. on an N core VM if JR is running `N-M` jobs then we can still run `M` jobs for other tests. With this python script we are going to loose this functionality. Also python this will fail for older release cycles ( where we do not have python3)

vlimant:
how about I put an "if python3 available run the new version" otherwise run whatever was before ?

smuzaffar:
> how about I put an "if python3 available run the new version" otherwise run whatever was before ?

still we loose the functionality of running JR jobs in parallel with other. Tail of JR jobs can waste resources which we could have used for other tests :-(

vlimant:
I understand better now the issue with the attempt in #1862

smuzaffar:
please test

comparisons/validateJR.py
```diff
@@ -0,0 +1,138 @@
...
+    processName = last_process(each_root_file)
+    if not processName in process_of_interest:
+        return
+    #print(f"found process of interest {processName} in file {each_root_file}")
+    path,fileName = each_root_file.rsplit('/',1)
+    ref_path =path.replace( options.base, options.ref )
+    _,fullName = path.rsplit('/',1)
+    wfn,therest=fullName.split('_',1)
+    wfn='wf'+wfn.replace('.','p')
+    fileindex = file_index(fileName)
```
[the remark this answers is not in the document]
vlimant: the logic is actually flawed as extra "+" are added, so not only separating steps.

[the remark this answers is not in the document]
vlimant: for example 10042.0_ZMM_13+2017+ZMM_13TeV_TuneCUETP8M1_GenSim+Digi+RecoFakeHLT+HARVESTFakeHLT+ALCA+Nano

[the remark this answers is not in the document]
vlimant: it's confusing because there are "+" in the directory names, that are not separating the workflows steps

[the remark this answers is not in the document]
vlimant: and I was assuming the directory name is <WF number>_<name of first step>+<name of second>+<...>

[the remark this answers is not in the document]
vlimant: I modified to `<dir name without +,_,.><file name without '.root'><wf number>`
makes rather long directory names though.

vlimant:
please test

comparisons/validateJR.py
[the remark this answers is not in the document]
slava77: > I modified to `<dir name without +,_,.><file name without '.root'><wf number>` makes rather long directory names though.

the cost of maintaining short-named map file that changes fairly infrequently (10 mins effort once a month or less) is not obviously larger than the extra cost of dealing with page-wide names (few? seconds 50-100 times per month)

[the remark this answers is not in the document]
slava77: > I see one less file to maintain, and the summary page anyways provide links to those with diff

sure, except that while automating it you probably put already around 3 years worth of effort of maintaining that file.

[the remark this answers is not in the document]
slava77: my wish would be to make the original workflow dir names short (80% is a waste there) and then for the comparisons we could just use those dir names directly and have all_OldVsNew for a top directory containing the wfs.
 I would likely not have invented the map file it if was the case.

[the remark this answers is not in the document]
vlimant: the short names can be defined directly in runTheMatrix, there actually is an unused field (except in the nano matrix) for each workflow to specify a keyword. Let us take note of this and provide short names all around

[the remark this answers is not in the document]
vlimant: an alternative https://github.com/vlimant/cms-bot/commit/f53a0dd6c90566dfff8d74165a6fcca55c501ce2 to only keep the mapping of the short names

vlimant:
while on naming @slava77 , naming the png using also the shortname is to be able to take the plot out and use it alone, while being able to identify the workflow from the png name, right ?

smuzaffar:
please test for CMSSW_12_1_X

last results look good, lets test it for other release cycles

smuzaffar:
please test for CMSSW_11_1_X

smuzaffar:
please test for CMSSW_9_4_X

smuzaffar:
please test for CMSSW_5_3_X

slava77:
> while on naming @slava77 , naming the png using also the shortname is to be able to take the plot out and use it alone, while being able to identify the workflow from the png name, right ?

I was thinking more of the directory names of what's in a relatively recent memory. But maybe a part of it is what you say. It could've even been picked up from your earlier version. I see this pattern in some archived logs since at least July 2012.

In retrospect, I don't think it's essential in the file name and we can have just the encoded variable there in the file name.

Also, I think the 1234.5 -> 1234p5 came up from lack of confidence in root detecting the format in the file name. We could get rid of it too and have the full dir name without re-parsing.

If I go along with your earlier proposal, the most straightforward to support could be `topDir/fullNameWorkflowDir/stepName/blah.png`
to be specific for 11634.0 from the recent IB:
`all_OldVSNew/11634.0_TTbar_14TeV+2021+TTbar_14TeV_TuneCP5_GenSim+Digi+RecoNano+HARVESTNano+ALCA/step3/recoMETs_tcMet__RECO_obj_phi.png`

without any transformations in the `fullNameWorkflowDir`.
And the way to fight with the long names would be from the workflow definition side

vlimant:
we can definitely make something like `topDir/fullNameWorkflowDir/stepName/blah.png`, but rather `topDir/fullNameWorkflowDir/processName_outputFileName/blah.png`.
I would not put `all` or `mini` in `topDir`, but still specify it to `validate` function ; which means topDir has no meaning. Hence `fullNameWorkflowDir/processName_outputFileName/blah.png` would be simpler.

logRootQA.py
```diff
@@ -205,12 +205,12 @@ def summaryJR(jrDir):
...
         nAll+=len(logs)
-        if len(logs)==1:
-            output=runCommand(['grep','DONE calling validate',logs[0]])
+        for log in logs:
+            output=runCommand(['grep','DONE calling validate',log])
             if len(output[0])>0:
                 nOK+=1
             else:
-                print('JR results failed',d)
+                print('JR results failed in ',log)
```
vlimant, line 213: this does not function with pr_testing/run-pr-comparisons `grep -a 'JR results failed' $QALOG | awk '{ split($0,a," "); print "<tr><td><a href=\"validateJR/"a[4]"\">"a[4]"</a></td><td align=\"right\"><a href=\"validateJR/"a[4]"/"a[4]".log\">log</a></td></tr>" }' >> $JRHTML`

smuzaffar:
please test with cms-sw/cmssw#40106

slava77:
> we can definitely make something like `topDir/fullNameWorkflowDir/stepName/blah.png`, but rather `topDir/fullNameWorkflowDir/processName_outputFileName/blah.png`. I would not put `all` or `mini` in `topDir`, but still specify it to `validate` function ; which means topDir has no meaning. Hence `fullNameWorkflowDir/processName_outputFileName/blah.png` would be simpler.

the job still needs to run somewhere and I think that it would be wrong to write into the runTheMatrix outputs. Hence the need for topDir. Also, in standalone running it's often convenient to have alternative selections (instead of `all`) or multiple iterations.

vlimant:
> > we can definitely make something like `topDir/fullNameWorkflowDir/stepName/blah.png`, but rather `topDir/fullNameWorkflowDir/processName_outputFileName/blah.png`. I would not put `all` or `mini` in `topDir`, but still specify it to `validate` function ; which means topDir has no meaning. Hence `fullNameWorkflowDir/processName_outputFileName/blah.png` would be simpler.
> 
> the job still needs to run somewhere and I think that it would be wrong to write into the runTheMatrix outputs. Hence the need for topDir. Also, in standalone running it's often convenient to have alternative selections (instead of `all`) or multiple iterations.

I think the latest version will work for you, as it seems to do this all, except maybe in standalone, if you run it from the matrix output directory, it will indeed put the subdirectories in the matrix output directories.

vlimant:
if the test runs :-(

vlimant:
@smuzaffar maybe we do not need 40106 for now in the test

vlimant:
N.B. the comparison did not run because the compilation of validate.C failed ... I rushed things and forgot basic tests ...

vlimant:
please test

vlimant:
all logs look good https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_12_6_X_2022-11-18-1100+6c4b56/54205/validateJR/ but no directory was copied back here

vlimant:
https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_12_6_X_2022-11-18-1100+6c4b56/54205/validateJR.html looks odd the file path is wrong

pr_testing/run-pr-comparisons
```diff
@@ -432,7 +433,7 @@ for f in $(find ${alt_dir} -maxdepth 1 -name 'command*' -o -name '*.ps' -o -name
 done
 
 #JR-comparison
-for x in $(find results/JR-comparison -maxdepth 2 -name 'command*' -o -name '*.log' -o -name '*.png' -type f) ; do
+for x in $(find results/JR-comparison -maxdepth 3 -name 'command*' -o -name '*.log' -o -name '*.png' -type f) ; do
```
vlimant, line 436: this will fix the copying of the output back to the web directory

logRootQA.py
```diff
@@ -193,24 +193,29 @@ def summaryJR(jrDir):
...
         break
 
     nAll=0
     nOK=0
-    for d in dirs:
-        diffs=getFiles(root+'/'+d,'*.png')
+    for d, subdir, files in os.walk(jrDir):
+        if not d.split('/')[-1].startswith('all_'): continue
+        if not '_' in d: continue
+        relative_d = d.replace(root,'')
```
vlimant, line 205: this should take care of printing the directory name relative to a root that gets changed consistently when copying to the web area

vlimant:
please test

slava77:
please test with cms-sw/cmssw#39808

to check with a PR that has more (expected) differences in many workflows

slava77:
> to check with a PR that has more (expected) differences in many workflows

here's a reference link to differences in cms-sw/cmssw#39808 https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_12_6_X_2022-11-18-1100+3763cc/54191/validateJR.html

slava77:
> **Summary**: https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-6c4b56/29134/summary.html
> **COMMIT**: [374e203](https://github.com/cms-sw/cms-bot/commit/374e203a7cb24427fad322aa8d56f636d833cb6e)
> **CMSSW**: CMSSW_12_6_X_2022-11-18-2300/el8_amd64_gcc10
> ****

the logs in comparisons and the outputs with diffs look reasonable with some spot-checking. 
I triggered tests to see a variant with more differences.

After that (or even in parallel?) a rerun in the older releases would be useful to confirm that things work there too

vlimant:
@slava77 39808 requires 40043, 40058, 40100 ?

slava77:
please test with cms-sw/cmssw#39808,cms-sw/cmssw#40043,cms-sw/cmssw#40058,cms-sw/cmssw#40100

slava77:
> Reco comparison results: 1267 differences found in the comparisons

vs reference
> Reco comparison results: 1108 differences found in the comparisons

looking at [the results in the new variant](https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_12_6_X_2022-11-19-1100+6c4b56/54213/validateJR.html), I see that 13234 and 13434 were incorrect in the map file (reference didn't cover); another reason to automate.

slava77:
please test for CMSSW_10_6_X

slava77:
please test for CMSSW_5_3_X

vlimant:
I need to make the changes to validate.C C++98 compatible for 5.3 ...

vlimant:
> In file included from /afs/cern.ch/work/v/vlimant/private/CMSSW_5_3_X_2022-11-20-0000/src/validate_lib/validate_C_ACLiC_dict.h:34:0,
>                  from /afs/cern.ch/work/v/vlimant/private/CMSSW_5_3_X_2022-11-20-0000/src/validate_lib/validate_C_ACLiC_dict.cxx:17:
> /afs/cern.ch/work/v/vlimant/private/CMSSW_5_3_X_2022-11-20-0000/src/validate_lib/./validate.C:33:6: warning: unused parameter 'step' [-Wunused-parameter]
> /afs/cern.ch/work/v/vlimant/private/CMSSW_5_3_X_2022-11-20-0000/src/validate_lib/./validate.C: In function 'PlotStats plotvar(TString, TString, bool)':
> /afs/cern.ch/work/v/vlimant/private/CMSSW_5_3_X_2022-11-20-0000/src/validate_lib/./validate.C:164:3: error: 'SetExponentOffset' is not a member of 'TGaxis'
> /afs/cern.ch/work/v/vlimant/private/CMSSW_5_3_X_2022-11-20-0000/src/validate_lib/./validate.C:165:3: error: 'SetExponentOffset' is not a member of 'TGaxis'
> In file included from /afs/cern.ch/work/v/vlimant/private/CMSSW_5_3_X_2022-11-20-0000/src/validate_lib/validate_C_ACLiC_dict.h:34:0,
>                  from /afs/cern.ch/work/v/vlimant/private/CMSSW_5_3_X_2022-11-20-0000/src/validate_lib/validate_C_ACLiC_dict.cxx:17:
> /afs/cern.ch/work/v/vlimant/private/CMSSW_5_3_X_2022-11-20-0000/src/validate_lib/./validate.C: At global scope:
> /afs/cern.ch/work/v/vlimant/private/CMSSW_5_3_X_2022-11-20-0000/src/validate_lib/./validate.C:1544:6: warning: unused parameter 'step' [-Wunused-parameter]
> /afs/cern.ch/work/v/vlimant/private/CMSSW_5_3_X_2022-11-20-0000/src/validate_lib/./validate.C: In function 'void validateEvents(TString, TString, TString, TString, bool, TString)':
> /afs/cern.ch/work/v/vlimant/private/CMSSW_5_3_X_2022-11-20-0000/src/validate_lib/./validate.C:1986:34: error: range-based 'for' loops are not allowed in C++98 mode
> /afs/cern.ch/work/v/vlimant/private/CMSSW_5_3_X_2022-11-20-0000/src/validate_lib/./validate.C:2848:34: error: range-based 'for' loops are not allowed in C++98 mode

those are not specific to this PR ; do we really have to go back to 5.3 ?

slava77:
> those are not specific to this PR ; do we really have to go back to 5.3 ?

is it newly broken or was it already broken?

vlimant:
I have started a test in 5.3 in https://github.com/cms-sw/cmssw/pull/40031

slava77:
> I have started a test in 5.3 in [cms-sw/cmssw#40031](https://github.com/cms-sw/cmssw/pull/40031)

I see there in the [tests from 12 days ago ](https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_5_3_X_2022-11-06-0000+3600ef/54007/validateJR/all_OldVSNew_TTbarFSpHLTwf5p1/all_OldVSNew_TTbarFSpHLTwf5p1.log)
```
Error: Function enable() is not defined in current scope  (tmpfile):1:
```
so, it was broken in some way already

vlimant:
indeed, https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_5_3_X_2022-11-06-0000+3600ef/54007/validateJR/validateJR.log the same compilation error is already there

slava77:
> indeed, https://cmssdt.cern.ch/SDT/jenkins-artifacts/baseLineComparisons/CMSSW_5_3_X_2022-11-06-0000+3600ef/54007/validateJR/validateJR.log the same compilation error is already there

perhaps it's an excuse to not go back that far in older releases ;)

slava77:
10_6_X results look OK.

slava77:
@clacaputo @mandrenguyen 
in case you haven't looked at this PR already , please check.
The format of what you are looking at frequently is going to change quite a bit

vlimant:
this https://github.com/vlimant/cms-bot/commit/fb036c73cad0712f97c497b9eebfc185441d3011 solves the compilation error in 5.3, but it does not avoid the error coming from
`FWLiteEnabler::enable()
Error: Function enable() is not defined in current scope  (tmpfile):1:
*** Interpreter error recovered ***`

smuzaffar:
For 5.3.X, [AutoLibraryLoader::enable()](https://github.com/cms-sw/cmssw/blob/CMSSW_5_3_X/FWCore/FWLite/interface/AutoLibraryLoader.h#L19)  is used. We call both `AutoLibraryLoader::enable()` and `FWLiteEnabler::enable()` to make sure that lib loadig is enable for any release. I think `FWLiteEnabler::enable() Error: Function enable() is not defined in current scope (tmpfile):1: *** Interpreter error recovered` error should not be causing any issues.

smuzaffar:
We can avoid `FWLiteEnabler::enable() Error: Function enable() is not defined in current scope (tmpfile)` if we conditionally add this if $CMSSW_RELEASE_BASE/rc/FWCore/FWLite/interface/FWLiteEnabler.h exists

smuzaffar:
+externals
As 5.3.X are already there which we can fix later. We can merge this after green single from @clacaputo @mandrenguyen

clacaputo:
+reconstruction

@vlimant thanks a lot for taking care of it

smuzaffar:
merging it now, @vlimant feel free to open a new PR for 5.3.X fixes

vlimant:
thanks !

PR #1877, opened 2022-11-07
[Jenkins parser] Handle `/cvmfs` not mounted on grid nodes

aandvalenzuela:
It includes a `create_gridnode` action and the error string "cannot open directory /cvmfs/.*cern.ch:" (it appears when trying to `ls` to a cvmfs repo and it is not mounted).
It also includes minor fixes and refactoring.


aandvalenzuela:
Merging since it is already running in Jenkins

PR #1882, opened 2022-11-15
[Parser] Add `retryLate` action to delay parser retries

aandvalenzuela:
This PR adds a new action `retryLate` that will delay the retry by 10 min. This is useful when the failure is due to server error, where an immediate retry will fail as well.

It also adds new failures and jobs to the config file.

The option for triggering the creation of grid nodes is disabled for debugging purposes.

### DIFF
--- a/jenkins/parser/actions.py
+++ b/jenkins/parser/actions.py
@@ -88,8 +88,8 @@ def trigger_retry_action(
         + str(regex.replace(" ", "&"))
         + '"'
     )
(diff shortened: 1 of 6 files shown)


aandvalenzuela:
Adding timestamp to retry exactly every 10 min

smuzaffar:
> Adding timestamp to retry exactly every 10 min

make it configurable ( as discussed)

PR #1899, opened 2022-12-04
Added E/Gamma L2

a-kapoor:
Please consider this PR to add me as the new E/Gamma L2


smuzaffar:
@swagata87 @lfinco any objections on this change (i.e replacing @lfinco with @a-kapoor for @cms-sw/egamma-pog-l2 )

PR #1901, opened 2022-12-07
Fixes and improvements in nodes-sanity-check job

aandvalenzuela:
This PR adds the option of testing one single lxplus node setting the boolean parameter `TEST_SINGLE_LXPLUS_HOST` to true. It also includes a special cleanup for non x86_64 (now it will only bring nodes online again if the offline reason is known, so in case one of the nodes is offline because of a different reason, e.g., building Spack, it won't bring the node online). Finally, `node-online` action on lxplus nodes brings the node online, but it stays connected to the buggy host. That is why we now send a `connect-node` action also so that jenkins runs the `connect.sh` script again avoiding the blacklisted host.


aandvalenzuela:
I have been testing these updates in Jenkins since 1st of December.

PR #1902, opened 2022-12-08
Update categories_map for #40264

mmusich:
Needed by https://github.com/cms-sw/cmssw/pull/40264


categories_map.py
```diff
@@ -541,7 +542,6 @@
     "DQM/SiPixelPhase1Config",
     "DQM/SiPixelPhase1DeadFEDChannels",
     "DQM/SiPixelPhase1Digis",
-    "DQM/SiPixelPhase1Heterogeneous",
```
makortel, line 544: The earlier package needs to be kept (the list should effectively be a union of all CMSSW releases).

PR #1907, opened 2022-12-15
Add wf `10804.31` and `10805.31` to known errors for release CMSSW_13_* also

aandvalenzuela:
We want to keep relvals wf  `10804.31` and `10805.31` as known errors in non-amd64 archs for the new  CMSSW_13_* release cycle. This will avoid showing TritonService server fallback errors in relvals as reported in https://github.com/cms-sw/cmssw/issues/37767.

### DIFF
--- a/cmssw_known_errors.py
+++ b/cmssw_known_errors.py
@@ -124,7 +124,7 @@ KNOWN_ERRORS["relvals"][RelFilter][".+"] = {
   "536.0": { "step": 1, "exitcode": 256,   "reason" : MSG_ASAN_INCOMPETIBILITY},
 }
 
-RelFilter="CMSSW_12_.+"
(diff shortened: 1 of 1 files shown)


aandvalenzuela:
@smuzaffar I think we forgot to include those relvals as known errors when changing the release cycle. They were already included for CMSSW_12_*

cmssw_known_errors.py
```diff
@@ -124,7 +124,7 @@
   "536.0": { "step": 1, "exitcode": 256,   "reason" : MSG_ASAN_INCOMPETIBILITY},
 }
 
-RelFilter="CMSSW_12_.+"
+RelFilter="CMSSW_(12|13)_.+"
```
smuzaffar, line 127: ah ok. I would suggest to change the regex so that it matches all releases 12 and above

Issue #1884, opened 2022-11-17
add automatic labeling of PR types from package lists


slava77:
to simplify searching/monitoring a list of relevant PRs, I'd like to introduce possibility to add lists of packages which, if modified in a PR will trigger a label for the PR.
This is instead of manual `type xyz`.

Opt-in strategy is probably OK.
This can start with a map file connecting type label with a list of packages.
Relevant `@cms-sw/xyz-{pog,dpg}-l2` would then be able to propose or signoff (if someone else decides to propose) on the acceptable lists.

smuzaffar:
@slava77 , so basically we extend https://github.com/cms-sw/cms-bot/blob/master/categories_map.py and add also a mapping for labels/packages e.g. something like [a]? And I guess the `label` in `categories_map.py` should have a matching `pog|dpg` entry in https://github.com/cms-sw/cms-bot/blob/master/categories.py  ... right?

[a]
```
CMSSW_LABELS = {
  "label": [
    "packages",
  ]
}
````

slava77:
> @slava77 , so basically we extend https://github.com/cms-sw/cms-bot/blob/master/categories_map.py and add also a mapping for labels/packages e.g. something like [a]? And I guess the `label` in `categories_map.py` should have a matching `pog|dpg` entry in https://github.com/cms-sw/cms-bot/blob/master/categories.py ... right?
> 
> [a]
> 
> ```
> CMSSW_LABELS = {
>   "label": [
>     "packages",
>   ]
> }
> ```

not exactly, or at least not at the level it's done in categories_map/categories py files.
I'm asking for a non-blocking label, while the examples you take are for blocking labels (e.g. `reconstruction-pending`).

As an example for tracking, I'd like to see for PRs touching packages in e.g. `RecoTracking` to be labeled with `tracking` (the same that I get now with `type tracking`). This is to simplify monitoring the relevant list of PRs. I would still rather have the blocking signatures like `assign tracking-pog` to remain manual.

The syntax can indeed be similar to what you quoted. One wish item, even for the current `categories_map.py` file is to accept wildcards or matching based on subsystem alone (not a full `subsystem/package` name).

smuzaffar:
[...]

PR #1914, opened 2023-01-17
Added support for non-blocking dpg/pog labels based on changed packages

smuzaffar:
Resolves https://github.com/cms-sw/cms-bot/issues/1884


process_pr.py
```diff
@@ -562,6 +569,12 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
             if package in category_packages:
                 package_categories[package].add(category)
                 pkg_categories.add(category)
+        for ex_lab, category_packages in list(CMSSW_LABELS.items()):
+            for cp in category_packages:
+                if package.startswith(cp):
```
slava77, line 574: is it possible to use regular expressions here instead of startswith?

PR #1916, opened 2023-01-18
add labeling for tracking

slava77:
add packages and subsystems to be labeled for tracking

- "CUDADataFormats/Track",
- "DataFormats/Track",
- "RecoPixelVertexing",
- "RecoTracker",
- "RecoVertex",
- "TrackPropagation",
    - will also match to TrackPropagation/SteppingHelixPropagator, which is mostly muon reco, but it's probably still OK for tracking
- "TrackingTools"
    - will also match TrackingTools/TrackAssociator, which is not tracking, but the alternative is to write a much longer list. So, we could probably unassign

@mmusich 
do you have any comments?

### DIFF
--- a/categories_map.py
+++ b/categories_map.py
@@ -1715,11 +1715,22 @@ CMSSW_CATEGORIES = {
 CMSSW_LABELS = {
   "tracking": [
     "CUDADataFormats/Track",
+    "CUDADataFormats/Vertex",
     "DataFormats/Track",
+    "DataFormats/VertexReco",
+    "DataFormats/V0Candidate",
+    "DataFormats/VZero",
+    "DQM/Tracking",
+    "FastSimulation/Tracking",
+    "Geometry/GlobalTracking",
     "RecoPixelVertexing",
     "RecoTracker",
     "RecoVertex",
     "TrackPropagation",
-    "TrackingTools"
+    "TrackingTools",
+    "Validation/RecoPixelVertexing",
+    "Validation/RecoTrack",
+    "Validation/RecoVertex",
+    "Validation/Tracking"
   ]
 }
--- a/categories_map.py
+++ b/categories_map.py
@@ -1711,9 +1711,9 @@ CMSSW_CATEGORIES = {
 #explicit `type label`. Only valid dpg/pog groups can be added as labels
 #Format: label = [cmssw_pkg1, cmssw_pkgN]
 #where cmssw_pkgX can be full or part of start of cmssw Package name
-# "jetmet" = [ "SubSystem1", "SubSystem2/Package"]
+# "jetmet": [ "SubSystem1", "SubSystem2/Package"]
 CMSSW_LABELS = {
-  "tracking" = [
+  "tracking": [
     "CUDADataFormats/Track",
     "DataFormats/Track",
     "RecoPixelVertexing",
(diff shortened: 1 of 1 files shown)


mmusich:
does it mean they will get labelled automatically?

categories_map.py
```diff
@@ -1711,6 +1711,15 @@
...
+# "jetmet": [ "SubSystem1", "SubSystem2/Package"]
 CMSSW_LABELS = {
+  "tracking": [
+    "CUDADataFormats/Track",
+    "DataFormats/Track",
+    "RecoPixelVertexing",
+    "RecoTracker",
+    "RecoVertex",
+    "TrackPropagation",
+    "TrackingTools"
```
mmusich, line 1723: Not sure if we need also `RecoPixelVZero` and `RecoVZero` (do not exist anymore since a long time).
Do we need to flag also non reconstruction packages?

slava77: > Do we need to flag also non reconstruction packages?

good point, here is more that matches

- CUDADataFormats/Vertex 
- DataFormats/VertexReco
- DQM/Tracking
- Validation/RecoPixelVertexing
- Validation/RecoTrack
- Validation/RecoVertex
- Validation/Tracking
- FastSimulation/Tracking
- Geometry/GlobalTracking 

what else?

slava77: > Not sure if we need also `RecoPixelVZero` and `RecoVZero` (do not exist anymore since a long time).

these were apparently removed in 9_2_X ( cms-sw/cmssw#18623) 10_0_X (cms-sw/cmssw#21456), respectively
But I guess I can add these as well

mmusich: I would add:
   * CondFormats/BeamSpotObjects
   * AnalysisDataFormats/TrackInfo 
   * CondCore/BeamSpotPlugins
   * CondFormats/BeamSpotObjects
   * CondTools/BeamSpot
   * DQM/BeamMonitor
   * Validation/TrackingMCTruth

slava77:
> does it mean they will get labelled automatically?

yes

categories_map.py
slava77: > I would add:

now in f2c36cc

mmusich:
looks good to me!

categories_map.py
slava77: actually in 9d8c766

mmusich:
there is tracking related code in `CommonTools/RecoAlgos` and possibly also scattered in other packages. I guess we can still flag by hand when needed.

slava77:
> there is tracking related code in `CommonTools/RecoAlgos` and possibly also scattered in other packages. I guess we can still flag by hand when needed.

right, I don't think there is a way to tag by individual files.

smuzaffar:
files information is available and we can use that for these labels.

mmusich:
> files information is available and we can use that for these labels.

doing a survey down to the file level now would be extremely tedious. I would suggest to continue incorporating files in the list once concrete requests materialize. What do you think?

smuzaffar:
> > files information is available and we can use that for these labels.
> 
> would this work already, or does it need some updates in the bot?

it need updates in bot code.

smuzaffar:
@slava77 , should I update bot to do the matching against the changed file ?

smuzaffar:
this has been merged and bot properly added the label for https://github.com/cms-sw/cmssw/pull/30373

PR #1917, opened 2023-01-19
update tracking-labeled packages and add trk-labeled packages

mmusich:
title says it all.
   * @slava77 FYI
   * @sroychow, this will automatically label tracker-dpg owned packages when a pull request to the central cmssw fork is made. Please let me know if you agree with the list.


sroychow:
@mmusich thanks! and I completely agree.

categories_map.py
```diff
@@ -1713,7 +1713,60 @@
...
+    "CondTools/SiPixel",
+    "CondTools/SiStrip",
+    "CUDADataFormats/SiPixel",
+    "CUDADataFormats/SiStrip",
+    "DPGAnalysis/SiStripTools",
+    "DataFormats/Phase2ITPixelCluster",
+    "DataFormats/Phase2Tracker",
+    "DataFormats/PixelMatchTrackReco",
+    "DataFormats/SiPixel",
+    "DataFormats/SiStrip",
```
slava77, line 1749: The `/SiStrip` and `/SiPixel` is repeated in many places.
It may be nice to have a possibility to match the package name based on just the package name with e.g. `/Package` to match (`startswith`) any subsystem? 
@smuzaffar do you think that would be practical or more problematic?

mmusich: in an earlier version I actually had `"DataFormats/Si"`, but fully specifying it looks more future proof.

slava77:
> @slava77 FYI

Thanks.

categories_map.py
smuzaffar: lets support regexp and bot converts all names to `.*(PackageName).*` which should allow you to use `/SiPixel` to match all packages with  `/SiPixel`

mmusich: does it require further change in the `cms-bot` code?
@smuzaffar let's merge this and revisit once it's available.

smuzaffar:
For non-blocking labels, bot now supports (https://github.com/cms-sw/cms-bot/pull/1918 ) regexp which are matched against file names instead of package names

categories_map.py
smuzaffar: bot is now ready to support regexp matched against file name.  If you want to update this PR then please go ahead and ping me when ready to go in

mmusich: I'd rather go as it is now.

mmusich:
@smuzaffar sorry for hi-jacking the thread here, but I couldn't find a better place to ask.
The non-blocking labels are great to flag ownership of a given PR, but they don't show in the release summary, e.g.: https://github.com/cms-sw/cmssw/releases/CMSSW_13_0_0_pre2 . Is there a way of making them visible ? It would greatly help the work of release validation.

smuzaffar:
@mmusich , sorry I missed this message. Sure, we can add these labels to the IB and release summary

PR #1919, opened 2023-01-19
add a trailing / in subsystem names for CMSSW_LABELS

slava77:
after #1918 the match became more inclusive than originally intended.
I added a trailing `/` to entries that initially corresponded to subsystem names

I'm not sure if a leading `^` is needed in each line.


slava77:
> I'm not sure if a leading `^` is needed in each line.

after some thought, adding `^` everywhere would be just more pedantic to match the original behavior pre-#1918 and verbose, which is not really required here

PR #1929, opened 2023-01-30
Add PerfTools/JeProf package under core signature

makortel:
Added in https://github.com/cms-sw/cmssw/pull/40624


gartung:
please test

PR #1935, opened 2023-02-08
Use github API to check/create CMSSW tag ...

iarspider:
... instead of cloning the cmssw repo; also prepare for code formatting with black


ib-create-tag.py
```diff
@@ -0,0 +1,105 @@
...
+
+    head = bran["commit"]
+    today = datetime.datetime.now().replace(
+        hour=HOUR, minute=0, second=0, microsecond=0
+    )
+    head_time = datetime.datetime.strptime(
+        head["commit"]["author"]["date"], "%Y-%m-%dT%H:%M:%SZ"
+    )
+
+    while head_time > today:
```
smuzaffar, line 80: @iarspider , why not use https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#list-commits and make use of `since` and `until` to get the commit before the hour?

ib-create-tag.py
```diff
@@ -0,0 +1,105 @@
...
+        ref = get_git_tag(repo, RELEASE_NAME)
+        OLD_HASH = ref["object"]["sha"]
+    except urllib.error.URLError:
+        OLD_HASH = "X"
+
+    head = bran["commit"]
+    today = datetime.datetime.now().replace(
+        hour=HOUR, minute=0, second=0, microsecond=0
+    )
+    head_time = datetime.datetime.strptime(
```
smuzaffar, line 76: Why `now` , it should be the day/time of the IB.

github_utils.py
```diff
@@ -578,3 +578,46 @@ def mark_commit_status(commit, repository, context="default", state="pending", u
...
+        "type": "commit",
+        "tagger": {"name": tagger_name, "email": tagger_email, "date": tag_date_s},
+    }
+    data = github_api("/repos/%s/git/tags" % repository, method="POST", params=params)
+    tag_sha = data["sha"]
+    params = {"ref": "refs/tags/%s" % tag_name, "sha": tag_sha}
+    github_api("/repos/%s/git/refs" % repository, method="POST", params=params)
+
+
+def get_commit_tags(repository, commit_sha, all_tags=False):
```
smuzaffar, line 616: did you look in to https://api.github.com/repos/cms-sw/cmssw/git/matching-refs/tags/CMSSW_12_6_X which returns all the tags matching `CMSSW_12_6_X` ? may be we can make use of that?

ib-create-tag.py
iarspider: Nice find

github_utils.py
iarspider: That is a good idea

smuzaffar:
@iarspider , is this ready?

iarspider:
@smuzaffar yes.

smuzaffar:
let me go through the changes before merging it.

github_utils.py
```diff
@@ -578,3 +578,70 @@ def mark_commit_status(commit, repository, context="default", state="pending", u
...
+    res = []
+    data = github_api("/repos/%s/tags" % repository, all_pages=all_tags, method="GET")
+    for tag in data:
+        if tag["commit"]["sha"] == commit_sha:
+            res.append(tag["name"])
+
+    return res
+
+
+def get_commits(repository, branch, until, per_page=1):
```
smuzaffar, line 627: @iarspider , did you check how many commints it returns ? May be we should also set `since` param

tag-and-schedule-ibs
```diff
@@ -72,21 +72,10 @@ for QUEUE in $RELEASE_QUEUES; do
...
     if [ "$CMSSW_TOOL_CONF_ONLY" != "true" ] ; then
-      git checkout $RELEASE_BRANCH
-      # find old tag, if any
-      OLD_HASH=`git show-ref $RELEASE_NAME | cut -f1 -d\ `
-      NEW_HASH=`git rev-list -n 1 --before="$DAY $HOUR:00" --first-parent $RELEASE_BRANCH`
-      # If there is no new hash, we have a problem.
-      # If there is no old hash, we simply tag.
-      # If there is an old hash, we use the old one.
-      if [ "X$NEW_HASH" = X ]; then
+      RELEASE_LIST="`./ib-create-tag.py -N ${RELEASE_NAME} -d $DAY -H $HOUR 11 -b ${RELEASE_BRANCH} -q ${QUEUE}`"
```
smuzaffar, line 75: ah looks like extea `11` here

github_utils.py
iarspider: I explicitly ask for only 1 commit (`per_page=1`)

smuzaffar: but this might not be the correct commit. For example for 11h00 IB, one could have merged a PR after 11h00 which contains a commit made at 10h58. This should return the commit matched by 
```
NEW_HASH=`git rev-list -n 1 --before="$DAY $HOUR:00" --first-parent $RELEASE_BRANCH`
```
right? So it should be the last merge commit on `$RELEASE_BRANCH` branch before the selected hour

iarspider: So I need to query 100 commits, and iterate until I find a merge commit, right?

iarspider: (100 or whatever max number of commits fit on one page)

ib-create-tag.py
```diff
@@ -0,0 +1,105 @@
...
+        OLD_HASH = "X"
+
+    commits_ = get_commits(repo, RELEASE_BRANCH, until=ib_date, per_page=100)
+    if not commits_:
+        exit(1)
+
+    head = None
+    for commit_ in commits_:
+        if commit_["commit"]["committer"]["name"] == "GitHub":
+            head = commit_
```
smuzaffar, line 83: @iarspider , should we `break` here?

smuzaffar:
@iarspider , any idea how to differentiate between merge commits like
- https://api.github.com/repos/cms-sw/cmssw/commits/fbcf9bc7cfdd7be9ab7dd9058d4205f3d3b5792c
  - this is a indirect merge commit via PR ( e.g. user opened a PR against his branch and then merge that)
- https://api.github.com/repos/cms-sw/cmssw/commits/c56dd8abcf2fea3b2ad8037028d8bdb4b58587a5
  - this is a valid cmssw merge commit

iarspider:
> @iarspider , any idea how to differentiate between merge commits like
> 
>     * https://api.github.com/repos/cms-sw/cmssw/commits/fbcf9bc7cfdd7be9ab7dd9058d4205f3d3b5792c
>       
>       * this is a indirect merge commit via PR ( e.g. user opened a PR against his branch and then merge that)
> 
>     * https://api.github.com/repos/cms-sw/cmssw/commits/c56dd8abcf2fea3b2ad8037028d8bdb4b58587a5
>       
>       * this is a valid cmssw merge commit

I guess we can check `commit.author.name` - it will be `cmsbuild` for the vaild merge commits and something else for indirect merges.

ib-create-tag.py
```diff
@@ -0,0 +1,106 @@
...
+    except urllib.error.URLError:
+        OLD_HASH = "X"
+
+    commits_ = get_commits(repo, RELEASE_BRANCH, until=ib_date, per_page=100)
+    if not commits_:
+        exit(1)
+
+    head = None
+    for commit_ in commits_:
+        if commit_["commit"]["committer"]["name"] == "GitHub" and commit_["commit"]["author"]["name"] == "cmsbuild":
```
smuzaffar, line 82: I am not sure if this is the right condition. Some time release managers merge the PR manually (because there were missing signatures and cmsbuild can not merge it). I am really not sure how to solve this issue, we really need to see the commit data and its parent

tag-and-schedule-ibs
```diff
@@ -72,21 +72,10 @@ for QUEUE in $RELEASE_QUEUES; do
...
-      OLD_HASH=`git show-ref $RELEASE_NAME | cut -f1 -d\ `
-      NEW_HASH=`git rev-list -n 1 --before="$DAY $HOUR:00" --first-parent $RELEASE_BRANCH`
-      # If there is no new hash, we have a problem.
-      # If there is no old hash, we simply tag.
-      # If there is an old hash, we use the old one.
-      if [ "X$NEW_HASH" = X ]; then
+      RELEASE_LIST="`./ib-create-tag.py -N ${RELEASE_NAME} -d $DAY -H $HOUR -b ${RELEASE_BRANCH} -q ${QUEUE}`"
+      if [ $? -ne 0 ]; then
         exit 1
       fi
```
smuzaffar, line 78: no need of the following block. The script has `-e` so script with fail and exit if `ib-create-tag.py ...` failed
```
if [ $? -ne 0 ]; then
        exit 1
fi
```

ib-create-tag.py
smuzaffar: @iarspider , i think we bit batter solution would be to add `from categories import CMSSW_L1` and change this condition to
```
if commit_["commit"]["committer"]["name"] == "GitHub" and commit_["commit"]["author"]["name"] in ["cmsbuild"]+CMSSW_L1:
```

smuzaffar:
@iarspider , https://github.com/cms-sw/cms-bot/pull/1935/files#diff-208ab6da7f86890a1ad6e8bf3df8497bb1301919d24b7205e9c1dbe25217c08aR26-R27 lines should be removed too

smuzaffar:
looks like something went wrong here

PR #1946, opened 2023-02-27
adding Simone Pigazzini as XPOG

vlimant:
FYI @simonepigazzini


simonepigazzini:
Somehow this is not active yet. For this https://github.com/cms-sw/cmssw/pull/41115, for instance, I was not tagged.

PR #1947, opened 2023-02-28
Use github api to check for/create CMSSW tag

iarspider:
Redo of  #1935

### DIFF
--- a/github_utils.py
+++ b/github_utils.py
@@ -605,6 +605,7 @@ def create_git_tag(repository, tag_name, commit_sha):
     tag_sha = data["sha"]
     params = {"ref": "refs/tags/%s" % tag_name, "sha": tag_sha}
     github_api("/repos/%s/git/refs" % repository, method="POST", params=params)
+    return tag_sha
 
 
 def get_commit_tags(repository, commit_sha, all_tags=False):
--- a/github_utils.py
+++ b/github_utils.py
@@ -595,17 +595,9 @@ def get_git_tag(repository, tag_name):
 
 def create_git_tag(repository, tag_name, commit_sha):
     get_gh_token(repository)
-    params = {
-        "tag": tag_name,
-        "message": "Create tag %s" % tag_name,
-        "object": commit_sha,
-        "type": "commit",
-    }
-    data = github_api("/repos/%s/git/tags" % repository, method="POST", params=params)
-    tag_sha = data["sha"]
-    params = {"ref": "refs/tags/%s" % tag_name, "sha": tag_sha}
+    params = {"ref": "refs/tags/%s" % tag_name, "sha": commit_sha}
     github_api("/repos/%s/git/refs" % repository, method="POST", params=params)
-    return tag_sha
+    return commit_sha
 
 
 def get_commit_tags(repository, commit_sha, all_tags=False):
--- a/github_utils.py
+++ b/github_utils.py
@@ -593,19 +593,13 @@ def get_git_tag(repository, tag_name):
     return data
 
 
-def create_git_tag(
-    repository, tag_name, commit_sha, tagger_name, tagger_email, tag_date=None
-):
+def create_git_tag(repository, tag_name, commit_sha):
     get_gh_token(repository)
-    if tag_date is None:
-        tag_date = datetime.datetime.now()
-    tag_date_s = tag_date.replace(microsecond=0).isoformat()
     params = {
         "tag": tag_name,
         "message": "Create tag %s" % tag_name,
         "object": commit_sha,
         "type": "commit",
-        "tagger": {"name": tagger_name, "email": tagger_email, "date": tag_date_s},
     }
     data = github_api("/repos/%s/git/tags" % repository, method="POST", params=params)
     tag_sha = data["sha"]
--- a/github_utils.py
+++ b/github_utils.py
@@ -497,7 +497,7 @@ def set_gh_user(user):
   global GH_USER
   GH_USER = user
 
-  
+
 def get_combined_statuses(commit, repository):
   get_gh_token(repository)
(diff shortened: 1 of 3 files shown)


smuzaffar:
looks good @iarspider  and mostly work. there are few things which we can still improve. See my comments in line

github_utils.py
```diff
@@ -578,3 +578,70 @@ def mark_commit_status(commit, repository, context="default", state="pending", u
...
+    get_gh_token(repository)
+    if tag_date is None:
+        tag_date = datetime.datetime.now()
+    tag_date_s = tag_date.replace(microsecond=0).isoformat()
+    params = {
+        "tag": tag_name,
+        "message": "Create tag %s" % tag_name,
+        "object": commit_sha,
+        "type": "commit",
+        "tagger": {"name": tagger_name, "email": tagger_email, "date": tag_date_s},
```
smuzaffar, line 608: `tagger` is optional , so so not explicitly pass it. The github token should should be enough to provide the needed information to github. So please remove tagger_name , tagger_email and drop `"tagger"` from the cal

ib-create-tag.py
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env python
+from __future__ import print_function
+
+import datetime
+import urllib.error
```
smuzaffar, line 5: this does not work for python2. I would suggest to use  
```
from _py2with3compatibility import HTTPError
```
and then replace
```
except urllib.error.URLError:
```
with
```
except HTTPError:
```

ib-create-tag.py
```diff
@@ -0,0 +1,108 @@
...
+            head = commit_
+            break
+
+    if head is None:
+        exit(1)
+
+    NEW_HASH = head["sha"]
+
+    if OLD_HASH == "X" and not opts.dryRun:
+        create_git_tag(
```
smuzaffar, line 94: we can save one api call to `get_commits` if `OLD_HASH != X"`  ... right? `commits_` are only needed it `OLD_HASH == "X"` , so may be the whole logic of getting the head commit can move under this condition

iarspider: @smuzaffar the script return early if there is already a tag `RELEASE_NAME` (and only return that tag) - right?

smuzaffar: you mean `tag-and-schedule-ibs` script ... right? I do not see that script returning if `RELEASE_NAME` exists ... may be I over looked ... can you please point me to that logic?

iarspider: I'm talking about your suggestion:
> we can save one api call to get_commits if OLD_HASH != X" ... right? commits_ are only needed it OLD_HASH == "X" , so may be the whole logic of getting the head commit can move under this condition

If we move the logic for getting head under `if OLD_HASH != X`, then should we exit early, returning just `RELEASE_NAME`,  or look for other tags that point to the same commit?

smuzaffar: looking for other tags is still needed, so no we can not return early.

ib-create-tag.py
```diff
@@ -0,0 +1,105 @@
...
+
+    repo = "%s/%s" % (GH_CMSSW_ORGANIZATION, GH_CMSSW_REPO)
+
+    try:
+        ref = get_git_tag(repo, RELEASE_NAME)
+        HEAD_SHA = ref["object"]["sha"]
+    except HTTPError:
+        commits_ = get_commits(repo, RELEASE_BRANCH, until=ib_date, per_page=100)
+        if not commits_:
+            exit(1)
```
smuzaffar, line 78: just for my education :-) I never used `exit()` but always imported the one from `sys` module. Do you have any preference on using `exit()` instead of `from sys import exit` ?

iarspider: No, no preference. And it seems that using `sys.exit` is a better approach: https://www.geeksforgeeks.org/python-exit-commands-quit-exit-sys-exit-and-os-_exit/

tag-and-schedule-ibs
```diff
@@ -72,21 +70,7 @@ for QUEUE in $RELEASE_QUEUES; do
...
-      # If there is an old hash, we use the old one.
-      if [ "X$NEW_HASH" = X ]; then
-        exit 1
-      fi
-      if [ "X$OLD_HASH" = X ]; then
-        git tag $RELEASE_NAME $NEW_HASH
-        git push origin --tags
-      fi
-      RELEASE_LIST="`git show --pretty='%d' HEAD | tr '[ ,()]' '[\n   ]'| { grep \"^${QUEUE}_20\" || true; }`"
+      RELEASE_LIST="`./ib-create-tag.py -N ${RELEASE_NAME} -d $DAY -H $HOUR -b ${RELEASE_BRANCH} -q ${QUEUE}`"
```
smuzaffar, line 73: ah tagging job failed, we should use `$CMS_BOT_DIR/ib-create-tag.py` instead of `./ib-create-tag.py`

ib-create-tag.py
```diff
@@ -0,0 +1,105 @@
...
+        if head is None:
+            sys.exit(1)
+
+        HEAD_SHA = head["sha"]
+        if not opts.dryRun:
+            create_git_tag(
+                repo,
+                RELEASE_NAME,
+                HEAD_SHA,
+            )
```
smuzaffar, line 97:
```suggestion
        if not opts.dryRun:
            create_git_tag(
                repo,
                RELEASE_NAME,
                head["sha"],
            )
        HEAD_SHA = get_git_tag(repo, RELEASE_NAME)["object"]["sha"]
```

something like this should fix the release list issue

iarspider: @smuzaffar I think we don't even need to do `get_git_tag` (at least when not dry-running) - we get tag sha back when creating a git tag.

smuzaffar: sounds correct

PR #1949, opened 2023-03-03
Disable root626 IB

iarspider:


smuzaffar:
thanks @iarspider for quick action, can you please also disable forward porting from master to rootnext branch?

iarspider:
@smuzaffar should I also disable forward porting to root628 branch?

smuzaffar:
> @smuzaffar should I also disable forward porting to root628 branch?

No root628 IBs we want to keep alive, so forward porting for that should be kept

PR #1963, opened 2023-03-31
override system commands for special IBs e.g ASAN

smuzaffar:
In ASAN IBs , some unit tests fail as `LD_PRELOAD=libasan.so` and some system command do not work well. For example `ps` commands hangs when  `LD_PRELOAD=libasan.so`  is set causing few tests to hangs for 3hours. This change allow the bot to conditionally override these system commands in such a way that it unset the LD_PRELOAD and then call system command.

hanging of `ps`  under `libasan` is known issue 
- https://gcc.gnu.org/bugzilla/show_bug.cgi?id=90589 
- https://sourceware.org/bugzilla/show_bug.cgi?id=27653


smuzaffar:
test parameters:
- full_cmssw = true
- release = CMSSW_13_1_ASAN_X

smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
+externals
looks good, unit tests in https://github.com/cms-sw/cmssw/pull/41248 passed

PR #1966, opened 2023-04-11
release-notes: look on more pages...

iarspider:
... if the requested release is not available on the first one

### DIFF
--- a/release-notes
+++ b/release-notes
@@ -138,7 +138,7 @@ if __name__ == "__main__":
       print("Modifying release notes for %s at %s" % (curr_release, url))
       print(urlopen(request, json.dumps({"body": header + cmssw_notes_str + cmsdist_header + cmsdist_notes_str })).read())
       print('ALL_OK')
-      return
+      exit(1)
 
   print("Release %s not found." % curr_release)
   exit(1)
--- a/release-notes
+++ b/release-notes
@@ -124,17 +124,21 @@ if __name__ == "__main__":
     print('ALL_OK')
     exit(0)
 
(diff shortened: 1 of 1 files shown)


iarspider:
There is a function to get list of releases in gh_utils, maybe we should use it?

release-notes
```diff
@@ -124,17 +124,21 @@ if __name__ == "__main__":
...
+    matchingRelease = [x["id"] for x in releases if x["name"] == curr_release]
+    if len(matchingRelease) > 0:
+      releaseId = matchingRelease[0]
+      url = "https://api.github.com/repos/cms-sw/cmssw/releases/%s" % releaseId
+      request = Request(url, headers={"Authorization" : "token " + token})
+      request.get_method = lambda: 'PATCH'
+      print("Modifying release notes for %s at %s" % (curr_release, url))
+      print(urlopen(request, json.dumps({"body": header + cmssw_notes_str + cmsdist_header + cmsdist_notes_str })).read())
+      print('ALL_OK')
+      return
```
smuzaffar, line 141: I think this will fail, you can not return from `for` loop

release-notes
```diff
@@ -124,17 +124,21 @@ if __name__ == "__main__":
...
+
+  print("Release %s not found." % curr_release)
+  exit(1)
```
smuzaffar, line 144: this means it will alway fail

smuzaffar:
for now this logic is fine ( i.e. iterating over pages) but I think we can just use https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#get-a-release-by-tag-name to get the release. All CMSSW releases are based on same tag. So we can just get the release by the tag name

PR #1969, opened 2023-04-17
Grant fwyzard the rights to sign any PR

fwyzard:
@dpiparo @jamesletts @perrotta @rappoccio 

I am officially requesting the rights to sign _any_ CMSSW pull request, to help with the migration of the HLT and Reconstruction code from CUDA to Alpaka.

### DIFF
--- a/categories.py
+++ b/categories.py
@@ -10,7 +10,7 @@ authors = {}
 #Any Githib user whose comments/requests should be ignored
 GITHUB_BLACKLIST_AUTHORS = []
 #CMS Offline Release Planning managers
-CMSSW_L1 = ["dpiparo", "perrotta", "rappoccio"]
+CMSSW_L1 = ["dpiparo", "perrotta", "rappoccio", "fwyzard"]
 #CMS-SDT members who has admin rights to various github organizations and repositories.
 #They are also reposionsible to sign for externals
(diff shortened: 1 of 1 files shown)


fwyzard:
please test

perrotta:
-1

@fwyzard do you intend to merge the PRs related to that migration bypassing the ORP? This is what the extra "power" you are requesting here would basically allow you, while I don't see any other practical application for it (well, you could also merge "any" PR and build "any" release, but I don't think this is what you are planning).

The design of the CMS software requires that the release management and release building are left in few dedicated hands, otherwise we risk to miss the connections between the different parts and put in danger the release for data taking (as well as the other ones). Therefore I reject this proposal of yours, while of course your suggestions are always welcome, and as we already did in the past we will be available to speed up the procedures for merging particular PRs based on your judgement and experience.

In any case, I am stepping down as release manager at the end of August and we are looking for a replacement: if you want to jump in, you are more than welcome.

fwyzard:
@perrotta, I do not feel it is necessary to merge PRs bypassing the release managers.

But the way the "signature" system works, requiring explicit signatures for trivial changes, _is_ a problem, when those signatures are not provided - like, for example, to a PR that is sitting idle for two weeks while simply adding a constant to a header file.

It seems in that case the only option is to bother the release managers explicitly and plead to bypass such signatures.

PR #1972, opened 2023-04-25
Add code to support triggering GPU unit-tests from bot

iarspider:


iarspider:
I have added the flag to [ib-run-pr-tests](https://cmssdt.cern.ch/jenkins/job/ib-run-pr-tests/) job, copied it and the logic to choose node from relvals one

smuzaffar:
@iarspider  , please revert the change for [ib-run-pr-tests](https://cmssdt.cern.ch/jenkins/job/ib-run-pr-tests/) job. We do not want to change that job otherwise we will not run non-GPU unit test. We want to run an extra job (which should be identical to https://cmssdt.cern.ch/jenkins/job/ib-run-pr-addon/ but sould run unit tests for gpu

smuzaffar:
-externals

smuzaffar:
we should add a new script `run-pr-unittests.sh` same as https://github.com/cms-sw/cms-bot/blob/master/pr_testing/run-pr-addon.sh but instead of running addon tests it should run `scram build unittests`

PR #1987, opened 2023-05-11
Shift log generator

iarspider:


iarspider:
@smuzaffar the code is ready, but I don't want to rewrite it to be py2-compatible. We can merge it despite test failures.

smuzaffar:
@iarspider , see if you can disable py2 tests for this otherwise once it is merged then py2 checks will always fail

iarspider:
@smuzaffar looks like you have already disabled them - https://github.com/cms-sw/cms-bot/commit/a4fd8e89789c98ba688997d7823aeb8efa5d0f5d

jenkins-jobs/git/git-notify-ib-updates
```diff
@@ -0,0 +1,215 @@
...
+    # noinspection PyUnresolvedReferences
+    from libib import PackageInfo, ErrorInfo
+except ImportError:
+    sys.path.insert(0, os.path.join(os.path.dirname(__file__), "..", ".."))
+    sys.path.insert(0, os.path.join(os.path.dirname(__file__), "..", "..", "shift"))
+    from github_utils import github_api
+    import libib  # NOQA
+
+    # noinspection PyUnresolvedReferences
+    from libib import PackageInfo, ErrorInfo
```
smuzaffar, line 25: @iarspider , I do not under the logic here. are then two `sys.path` at line number 10 and 11 enough? why you have to insert the same path again?

jenkins-jobs/git/git-notify-ib-updates
```diff
@@ -0,0 +1,215 @@
...
+try:
+    current_shifter = libib.fetch("/SDT/shifter.txt", content_type=libib.ContentType.TEXT)
+    exit_code = 0
+except (urllib.error.URLError, urllib.error.HTTPError) as e:
+    print("WARNING: failed to get current_shifter!")
+    print(e)
+    current_shifter = "all"
+    exit_code = 1
+
+cache_root = "/build/builds/cms-ib-notifier"
```
smuzaffar, line 36: why hard coded path here? I guess it is path on jenkins controller node ... right? If yes then I guess you can use env `JENKINS_HOME` ( so the full path will be `${JENKINS_HOME}/builds/cms-ib-notifier` ) . Looks like this is a cache directory which is persistant on controller, in tis case I would suggest to use `${JENKINS_HOME}/workspace/cache/cms-ib-notifier` .

jenkins-jobs/git/git-notify-ib-updates
```diff
@@ -0,0 +1,215 @@
...
+        uri="/repos/{}/commits/{}".format(repo, commit), method="GET", all_pages=False
+    )
+
+
+def isoparse(strDate):
+    return datetime.datetime.strptime(strDate, "%Y-%m-%dT%H:%M:%SZ")
+
+
+def main():
+    workspace = os.getenv("WORKSPACE", os.path.expanduser("~"))
```
smuzaffar, line 59: I would suggest to use pwd as default instead of HOME. HOME on jenkins controller is `/var/lib/jenkins` and we do not want to write any thing in that path

jenkins-jobs/git/git-notify-ib-updates
iarspider: This path was suggested by Andrea. I will change it following your recommendations.

jenkins-jobs/git/git-notify-ib-updates
iarspider: Ok

jenkins-jobs/git/git-notify-ib-updates
iarspider: @smuzaffar these lines were for local testing. Cleaned up.

smuzaffar:
looks good to me. I am merging it now

PR #2007, opened 2023-06-23
[PARSER] Add compare-root-files-short-matrix job

aandvalenzuela:


aandvalenzuela:
There is no need to further testing (sanity checks passed)

PR #2012, opened 2023-07-07
Allow back "operations" signature for perrotta

perrotta:
Having added a line in which I can sign for "alca" and "db" had overwritten my "operations" category. This should restore it (and I'll remove as soon as I will not be ORP any more)


francescobrivio:
ooops sorry Andrea I didn't think of that! 😅

PR #2022, opened 2023-07-20
Build ROOTDBG IB on-demand only

aandvalenzuela:
Removing `BUILD_HOUR=23` and adding `NO_IB=1` to build `ROOTDBG_X` on-demand


smuzaffar:
thanks @aandvalenzuela . As we now have ppc64le IB with root debug, can you please inform vassil so that he can test it on ibmminsky nodes

PR #2011, opened 2023-07-03
Add D98 prodlike wf for profiling

srimanob:
Add D98 prodlike wf for profiling

PR #2025, opened 2023-07-31
Add D98 wf to profile runner

mandrenguyen:
The phase 2 D98 workflow was added to `cmssw-pr-test-config` in #2011 
The PR adds it also to `profileRunner.py`

### DIFF
--- a/reco_profiling/profileRunner.py
+++ b/reco_profiling/profileRunner.py
@@ -162,6 +162,24 @@ workflow_configs = {
         "nThreads": 1,
         "matrix": "upgrade"
     } ,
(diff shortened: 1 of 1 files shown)


smuzaffar:
enable profiling

smuzaffar:
please test

mandrenguyen:
@smuzaffar it says the profiling is still running, but that doesn't appear to be the case.
It looks to me that the profiling and unittests are failing only b/c of DAS issues.  
Do you see something different?

smuzaffar:
+externals

profiling job for `24834.21` ran fine

PR #2031, opened 2023-08-09
Adding "node offline" as reason for a job to be in the queue

aandvalenzuela:
Already tested locally.


aandvalenzuela:
Changes are visible in OpenSearch

PR #2033, opened 2023-08-09
Update categories.py for pdmv

perrotta:
Add Adriano and Mikko to the pdmv category


smuzaffar:
+externals
looks good

PR #2036, opened 2023-08-15
Make sure all scripts can run without python2

iarspider:


smuzaffar:
@iarspider , for now, do not touch DMWM files

smuzaffar:
we will never be able to merge and test all these changes :-(

PR #2039, opened 2023-08-18
Monitor jenkins queue and offline nodes

aandvalenzuela:
Hello,

This script monitors the jenkins queue and looks for the buggy offline nodes (i.e., the nodes that are kept offline in Jenkins and we need to manually launch them).

We are seeing this type of errors more often, so I think this monitoring will help in detecting the nodes.

As discussed, there is a fist job that runs a simple check daily on every node (https://cmssdt.cern.ch/jenkins/view/All/job/jenkins-infrastructure-sanity-check/).
Then the monitoring job (https://cmssdt.cern.ch/jenkins/view/All/job/jenkins-monitor-queue/) runs every hour and reports offline nodes to OpenSearch. Offline buggy nodes can be found in the Jenkins Grafana Dashboard (https://monit-grafana.cern.ch/d/83t0M504k/cmssdt-jenkins?orgId=11) with the total offline time (in minutes), if there is any buggy node.

### DIFF
--- a/jenkins_monitor_queue.py
+++ b/jenkins_monitor_queue.py
@@ -1,0 +1,84 @@
+#!/usr/bin/env python
+
+import datetime, json
+from hashlib import sha1
+from es_utils import get_payload_wscroll, send_payload, delete_hit
+
+JENKINS_PREFIX="jenkins"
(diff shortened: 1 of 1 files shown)


aandvalenzuela:
I get it in to continue testing. It has been running for a couple of days in Jenkins and it does not affect any other jobs

PR #2040, opened 2023-08-18
Get failing relvals from OpenSearch

aandvalenzuela:
It goes along with https://github.com/cms-sw/cms-docker/pull/210.
It gets the relval failures from OpenSearch. This information is used to not fail jobs if relvals are failing also in IBs.

### DIFF
--- a/get-relval-failures.py
+++ b/get-relval-failures.py
@@ -1,0 +1,50 @@
+#!/usr/bin/env python3
+
+import argparse
+from es_utils import get_payload_wscroll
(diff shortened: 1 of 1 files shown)


smuzaffar:
I have updated the script. The output format is `WF:<wf-num>:step<n>:<exitcode>` and then cmssw test script should just grep `WF:<wf-num>:step<n>:<exitcode>` in this output and if exact match is found then that wf/step is known failure in Ib

PR #2044, opened 2023-08-23
Look for reco event loop and fast timer reports in any arch subdir.

gartung:
The release ARCHITECTURE is changed from el8_* to slc7_* in the profiling job to allow accurate memory profiling.


gartung:
@smuzaffar

smuzaffar:
@gartung , note that slc7 arch will be dropped by the in Nov/Dev this year.

gartung:
I decided to only run the `ib-run-igprof` Jenkins jobs in an cc7 image

PR #2046, opened 2023-08-28
Py3: More changes

iarspider:
`process_pr` fix is just to remove error (one should always use `open` to open a file). There are more changes required to make `process_pr` fully py3-compatible: I see many calls to `.encode("ascii", "ignore")` and comparision of result with string. This will no longer work, since `encode` returns `bytes` object, which can't be compared to `str` (directly or with a regex). Do you remember why these calls were added?

### DIFF
--- a/jenkins_callback.py
+++ b/jenkins_callback.py
@@ -26,7 +26,7 @@ def build_jobs(jenkins_url, jobs_data, headers = {}, user="cmssdt"):
              "Submit": "Build"
            }
     try:
-      data = urlencode(data)
+      data = urlencode(data).encode()
       req = Request(url=url,data=data,headers=headers)
       content = urlopen(req).read()
       print ("ALL_OK")
--- a/trigger_jenkins_job.py
+++ b/trigger_jenkins_job.py
@@ -1,5 +1,4 @@
-#!/usr/bin/env python
-from __future__ import print_function
+#!/usr/bin/env python3
 from optparse import OptionParser
 from jenkins_callback import build_jobs
 import json
--- a/add-externals-gh-labels.py
+++ b/add-externals-gh-labels.py
@@ -1,5 +1,4 @@
-#!/usr/bin/env python
-from __future__ import print_function
+#!/usr/bin/env python3
 from github import Github
 from os.path import expanduser, dirname, abspath, join, exists
 from githublabels import LABEL_TYPES, COMMON_LABELS, COMPARISON_LABELS, CMSSW_BUILD_LABELS, LABEL_COLORS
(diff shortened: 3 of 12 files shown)


iarspider:
Ready for review

smuzaffar:
>I see many calls to .encode("ascii", "ignore") and comparision of result with string. This will no longer work, since encode returns bytes object, which can't be compared to str (directly or with a regex). 

yes , github user name can contain non-ascii chars so one need encoding

smuzaffar:
are all these formatting changes needed for py3? If not then please remove them , this will help with the review.

iarspider:
@smuzaffar I have tested most of the files touched (except `es_ibs_log.py`, which requires log files, and `process_pr.py`, which requires a lot more changes), should be fine.

jenkins_callback.py
```diff
@@ -26,7 +26,7 @@ def build_jobs(jenkins_url, jobs_data, headers = {}, user="cmssdt"):
              "Submit": "Build"
            }
     try:
-      data = urlencode(data)
+      data = urlencode(data).encode()
```
smuzaffar, line 29: @iarspider , does this work for both python2 and python3. This code is used by cmssdt-web too ( https://github.com/search?q=repo%3Acms-sw%2Fcmssdt-web%20jenkins_callback&type=code )

trigger_jenkins_job.py
```diff
@@ -1,5 +1,4 @@
-#!/usr/bin/env python
-from __future__ import print_function
+#!/usr/bin/env python3
```
smuzaffar, line 1: were you able to trigger any jenkins job using this?

smuzaffar: looks fine, I was able to trigger https://cmssdt.cern.ch/jenkins/view/Jenkins%20Tests/job/jenkins-test-job3/1/console

jenkins_callback.py
smuzaffar: yes looks good

smuzaffar:
@iarspider , for `es_ibs_logs.py` I am getting errors like
```
Working on  /data/sdt/SDT/jenkins-artifacts/HLT-Validation/CMSSW_13_3_X_2023-08-29-1100/el8_amd64_gcc11/OnLine_HLT_GRun_DATA.log
  File "./es_ibs_log.py", line 237, in <module>
    process_hlt_log(logFile)
  File "./es_ibs_log.py", line 118, in process_hlt_log
    for index, l in enumerate(f):
  File "/usr/lib64/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 3867: ordinal not in range(128)
```

smuzaffar:
+externals
sample run of tests look good, lets get this in.

PR #2048, opened 2023-08-28
Continue after segfault in igprof so igprof databases are generated. Remove use of tee.

gartung:


gartung:
@smuzaffar

gartung:
please test

gartung:
please test

PR #2060, opened 2023-08-31
add Javier Fernandez as reco convener

mandrenguyen:
- Adding Javier starting Sept 1st
- Setting Claudio's end date to Oct 1st (just to give a bit of margin)


smuzaffar:
@mandrenguyen , `cmssw_l2/l2.json` is auto generated from the content of `categories.py`. So please revert the change for l2.json and just remove @clacaputo too in `categories.py`.

PR #2064, opened 2023-09-01
Switch Phase-2 wf to 25034.21

mandrenguyen:
For the new D98 geometry we accidentally used 24834.21 which is without PU.
This PR switches to the correct 200 PU workflow for profiling.


smuzaffar:
please test

mandrenguyen:
gentle ping

PR #2066, opened 2023-09-11
Found some more scripts that can/should be updated to Py3

iarspider:
I think only `github_scripts/simultaneous_files_modifications_by_PRs.py` is actually used, but could be wrong.

### DIFF
--- a/create-new-data-pr.py
+++ b/create-new-data-pr.py
@@ -4,11 +4,10 @@ from github import Github
 from optparse import OptionParser
 import repo_config
 from os.path import expanduser
-from urllib2 import urlopen
 from json import loads
 import re
 from github_utils import get_git_tree
-from _py2with3compatibility import run_cmd, HTTPError
+from _py2with3compatibility import run_cmd, HTTPError, urlopen
 
 def update_tag_version(current_version):
     updated_version = int(current_version)+1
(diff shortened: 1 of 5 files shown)


create-new-data-pr.py
```diff
@@ -68,7 +67,10 @@ def get_tag_from_string(tag_string=None):
   # if created files and modified files are the same count, all files are new
 
   response = urlopen("https://api.github.com/repos/%s/pulls/%s" % (opts.data_repo, opts.pull_request))
-  res_json = loads(response.read())
+  res = response.read()
+  if isinstance(res, bytes):
+    res = res.decode()
+  res_json = loads(res)
```
smuzaffar, line 73: @iarspider , do we need this is script runs under python3?
by the way, you need to update the `shebang` to use python3 ... right?

iarspider: This script is used by https://cmssdt.cern.ch/jenkins/job/update-data-tag-on-pr-merge/, which runs on cmssdt nodes. I will update hashbang, thanks for noticing.

smuzaffar: I mean do we need this logic https://github.com/cms-sw/cms-bot/pull/2066/files#diff-d401cb5613b47d8eca73383d19d96514e9af6681747d2bcd461c4d6812c334aaR70-R73 if script only runs under python3?

iarspider:
* create-gh-pr.py: used by https://cmssdt.cern.ch/jenkins/job/GitHub-refactor-cmssw-module (not ran since 2020, triggered by https://cmssdt.cern.ch/jenkins/job/GitHub-refactor-cmssw/) and https://cmssdt.cern.ch/jenkins/job/process-deprecate-releases/ (nor ran since 2016)
* create-gh-release.py: unused?
* create-new-data-pr.py: used by https://cmssdt.cern.ch/jenkins/job/update-data-tag-on-pr-merge/
* github_scripts/simultaneous_files_modifications_by_PRs.py:used by https://cmssdt.cern.ch/jenkins/job/run-pr-code-checks/ (via `run-pr-code-checks` script)
* modify_comment.py: wrapper around function with the same name in `process_pr.py`, referenced by (unused?) `run-user-pr-test`

create-new-data-pr.py
iarspider: Technically, no: `urlopen`, I think, now always return `bytes`.

smuzaffar:
please test for CMSSW_5_3_X

PR #2067, opened 2023-09-12
add  DQMOffline/RecoB to tracking (non-blocking) category

slava77:
in categories_map.py

### DIFF
--- a/categories_map.py
+++ b/categories_map.py
@@ -1811,6 +1811,7 @@ CMSSW_LABELS = {
     "DataFormats/V0Candidate",
     "DataFormats/VZero",
     "DQM/BeamMonitor",
+    "DQMOffline/RecoB",
     "DQM/Tracking",
     "FastSimulation/Tracking",
     "Geometry/GlobalTracking",


mmusich:
I was wondering if instead of this, we should actually move out [PrimaryVertexMonitor](https://github.com/cms-sw/cmssw/blob/6d2f66057131baacc2fcbdd203588c41c885b42c/DQMOffline/RecoB/plugins/PrimaryVertexMonitor.cc) out of `DQMOffline/RecoB` (which looks mostly for BTV usage) into a different folder (as far as I can tell we don't have a dedicated Vertexing DQM subsystem, but we could create a `DQM/VertexingMonitor` package and then move in it other vertexing related plugins scattered around).

smuzaffar:
@slava77 @mmusich , let me know if you want to update this. I am fine with this change too (at least this will work for old release cycles)

mmusich:
>  I am fine with this change too (at least this will work for old release cycles)

OK, that's a good point. Let's go with this then.

PR #2078, opened 2023-09-26
Add generation of property file for creating data repo

iarspider:


process_pr.py
```diff
@@ -1,7 +1,10 @@
+import os
+
```
smuzaffar, line 2: ?? why

process_pr.py
```diff
@@ -684,6 +687,10 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
         signing_categories.add("tests")
         push_test_issue = True
     except: pass
+    if (not issue.pull_request) and re.match(CREATE_REPO, issue.title):
```
smuzaffar, line 690: Thsi section is only run for issues so no need to check for `not issue.pull_request`. Replace `not issue.pull_request` with `cmssw_repo` ( property file should be createdonly for cmssw issues)

process_pr.py
```diff
@@ -684,6 +687,10 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
+      with open("query-new-data-repo-issues.properties", "w") as f:
+        f.write("CMS_BOT_BRANCH="+os.environ.get("CMS_BOT_BRANCH", "master"))
```
smuzaffar, line 692: remove this. We want to use the default branch set in the jenkins job

process_pr.py
```diff
@@ -684,6 +687,10 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
+        f.write("ISSUE_NUMBER="+str(issue.number))
```
smuzaffar, line 693: I guess you are missing `newline` char here

process_pr.py
```diff
@@ -684,6 +687,10 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
         signing_categories.add("tests")
         push_test_issue = True
     except: pass
+    if (not issue.pull_request) and re.match(CREATE_REPO, issue.title):
+      with open("query-new-data-repo-issues.properties", "w") as f:
```
smuzaffar, line 691: add the issue number in the property file name so that for each issue it creates a unique file

PR #2079, opened 2023-09-26
Use el8 buildtime images for DEVEL (amd64) and ROOTDBG (ppc64le) IBs

aandvalenzuela:
@smuzaffar, should I set an dummy IB (based on DEFAULT `el8_aarch64_gcc11`) to test aarch64?


smuzaffar:
yes @aandvalenzuela  add CMSSW_13_3_ROOTDBG_X for aarch64bto test the container

PR #2105, opened 2023-11-02
Do not reset all labels on commit

iarspider:


process_pr.py
```diff
@@ -1110,6 +1114,47 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
+                    ):
+                        test_comment = comment
+                    if sign == "orp":
+                        mustClose = False
+            elif ctype == "-1":
+                for sign in selected_cats:
+                    signatures[sign] = "rejected"
+                    if sign == "orp":
+                        mustClose = False
+            # continue
```
iarspider, line 1146: This `continue` was not, I think, needed originally - this block was the last one in the "for each comment" loop...

iarspider:
This requires much deeper modifications to process_pr.py, will create a new PR once ready.

PR #2106, opened 2023-11-06
Allow manual override to PR test results

iarspider:


iarspider:
Resolves cms-sw/cmssw#43143

process_pr.py
```diff
@@ -1110,6 +1114,36 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
+                    ):
+                        test_comment = comment
+                    if sign == "orp":
+                        mustClose = False
+            elif ctype == "-1":
+                for sign in selected_cats:
+                    signatures[sign] = "rejected"
+                    if sign == "orp":
+                        mustClose = False
+
```
smuzaffar, line 1146: @iarspider , I think moving this block of code above the `comment.created_at < last_commit_date` check is wrong. This means that a `+1` by a l2 is going to approve all the future changes too ... right?

iarspider: I keep forgetting that we don't reset the signatures, but actually set them.

process_pr.py
```diff
@@ -1237,6 +1242,16 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
+                elif REGEX_TEST_IGNORE.match(first_line):
+                    reason = REGEX_TEST_IGNORE.match(first_line)[1].strip()
+                    if reason not in TEST_IGNORE_REASON:
+                        print("Invalid ignore reason:", reason)
+                        set_comment_emoji(comment.id, repository, "-1")
+                        reason = ""
+
+                    if reason:
+                        override_tests_failure = reason
+                        set_comment_emoji(comment.id, repository)
```
smuzaffar, line 1254: @iarspider , please call these  `set_comment_emoji()`  only if not `dryRun`

process_pr.py
```diff
@@ -1501,6 +1514,13 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
             l = cat + "-" + signatures[cat]
         labels.append(l)
 
+    # Process "ignore tests failure"
+    if override_tests_failure and signatures["tests"] == "rejected":
+        labels.remove("tests-rejected")
+        labels.append("tests-approved")
+        labels.append("tests-" + override_tests_failure)
+        signatures["tests"] = "approved"
+
```
smuzaffar, line 1523: why not move this logic bit above e.g.
```
# Labels coming from signature.
labels = []
if override_tests_failure and signatures["tests"] == "rejected":
    signatures["tests"] = "approved"
    labels.append("tests-" + override_tests_failure)
for cat in signing_categories:
....
....
```

process_pr.py
```diff
@@ -1559,6 +1579,10 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
     xlabs = ["backport", "urgent", "backport-ok", "compilation-warnings"]
     for lab in TYPE_COMMANDS:
         xlabs.append(lab)
+
+    if set(labels).intersection(set("tests-" + x for x in TEST_IGNORE_REASON)):
+        labels.append("tests-approved")
+
```
smuzaffar, line 1585: do we need this? aren't we adding `tests-approved` in the changes above

process_pr.py
iarspider: No particular reason. Will do.

process_pr.py
```diff
@@ -1732,6 +1756,8 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
         requiresTestMessage = " (tests are also fine)"
     elif "tests-rejected" in labels:
         requiresTestMessage = " (but tests are reportedly failing)"
+    elif labels.intersection(set("tests-" + x for x in TEST_IGNORE_REASON)):
```
smuzaffar, line 1759: it should be `set(labels)` here ... right?

smuzaffar: by the way , this condition will not hit as above we have `if "tests-approved" in labels:`. May be change it to 

```
if labels.intersection(set("tests-" + x for x in TEST_IGNORE_REASON)):
  requiresTestMessage = " (test failures were overridden)"
elif "tests-approved" in labels:
...
...
```

iarspider: > it should be `set(labels)` here ... right?

 `labels` is already `set` here  https://github.com/cms-sw/cms-bot/blob/ff41a6536c621196cca6769e02559dfc09885d46/process_pr.py#L1607

smuzaffar:
looks good to go in

PR #2110, opened 2023-11-09
Do not reset all labels on commit

iarspider:


pyproject.toml
```diff
@@ -0,0 +1,547 @@
+[tool.black]
```
smuzaffar, line 1: @iarspider , this has nothing to do with the bot changes ..... right? I would suggest to open a separate PR for this. By the way what is the purpose of this file?

process_pr.py
```diff
@@ -1003,6 +1036,9 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
                     pull_request_updated = True
             continue
 
+        if (commenter == cmsbuild_user) and re.match(CMSBOT_TECHNICAL_MSG, first_line):
```
smuzaffar, line 1039: also check if `technical_comment` is not already set. may be merge this block with the one above

process_pr.py
```diff
@@ -1255,38 +1325,74 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
-                    signatures[sign] = "rejected"
-                    if sign == "orp":
-                        mustClose = False
-
     # end of parsing comments section
 
+    # Get the commit cache from `already_seen` commit or technical commit
+    print("Recalculating signatures")
+    cache_comment = None
+    new_comment = False
```
smuzaffar, line 1333: `new_comment` is not needed. why not just use `cache_comment`

process_pr.py
```diff
@@ -1255,38 +1325,74 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
+    old_body = cache_comment.body if cache_comment else CMSBOT_TECHNICAL_MSG
+    new_body = REGEX_COMMITS_SEEN.sub(
+        "<!-- commits cache: " + dumps_compact(commit_cache) + " -->", old_body
+    )
+
+    if not dryRun:
+        if new_comment:
+            issue.create_comment(new_body)
+        else:
+            cache_comment.edit(new_body)
```
smuzaffar, line 1360: only do it if `old_body != new_body`

process_pr.py
```diff
@@ -1255,38 +1325,74 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
+            print("DRY RUN: Updating existing comment with text")
+            print(new_body)
+
+    for commit in all_commits:
+        if commit.sha not in commit_cache:
+            commit_cache[commit.sha] = {
+                "time": int(commit.commit.committer.date.timestamp()),
+                "files": get_changed_files_in_commit(repo, commit),
+            }
+
```
smuzaffar, line 1375: I think creating/updating the cache_comment should be done after this.

smuzaffar: for each commit we are making one api call also we make one api call to get all the files changes here https://github.com/cms-sw/cms-bot/pull/2110/files#diff-f37d46a488510bd0b96fd201ff189f8a704b756502df6a1cd46b117986e608a4L772 . We need need to make all files changed call? Can the logic for file->package->category be move down here (this will save us one extra pi call)

smuzaffar: also we should add some protection against the PRs like https://github.com/cms-sw/cmssw/pull/38975 ( where someone used the wrong branch to open PR and it contains over 1000 commits). We can not afford to waste api calls for such PRs. I think for newly only PRs if commit count is above a threshold then bot should not process commits instead it should either add a comment to inform user about it. We can add a bot command to override this if a PR really contains commits more than default threshold

pyproject.toml
iarspider: Sure, I will split it out. In general, this file describes how a Python package is built. In our case, it contains configuration for Black (clang-format for Python) and pylint (clang-tidy) - with this file you can just run `black --check <filename>` instead of `black --check -l 99 -t py36 .... py311`.

process_pr.py
iarspider: Will do

process_pr.py
iarspider: Will look into it.

process_pr.py
iarspider: Will look into it.

process_pr.py
iarspider: Will look into it.

iarspider:
NB: Have not yet tested last changes properly.

process_pr.py
```diff
@@ -1254,38 +1348,117 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
+    if cache_comment:
+        seen_commits_match = REGEX_COMMITS_CACHE.search(cache_comment.body)
+        if seen_commits_match:
+            print("Loading commit cache")
+            commit_cache = loads(seen_commits_match[1])
+
+    if pr.commits < MAX_INITIAL_COMMITS_IN_PR or ok_too_many_commits:
+        for commit in all_commits:
+            if commit.sha not in commit_cache:
+                commit_cache[commit.sha] = {
```
smuzaffar, line 1385: @iarspider , we should call `api_rate_limits_repo` here to slow down the processing if api rate is reaching to 0

process_pr.py
```diff
@@ -1003,6 +1014,21 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
+            (commenter == cmsbuild_user)
+            and re.match(CMSBOT_TECHNICAL_MSG, first_line)
+            and not technical_comment
+        ):
+            technical_comment = comment
+
+        if (commenter == cmsbuild_user) and "This PR contains too many commits" in first_line:
+            warned_too_many_commits = True
+
+        if commenter in l2s and re.match(
```
smuzaffar, line 1027: @iarspider , to match the exact github user I think it is better to use `commenter in CMSSW_ISSUES_TRACKERS` instead of ` commenter in l2s`

process_pr.py
```diff
@@ -740,6 +738,13 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
     # For future pre_checks
     # if prId>=somePRNumber: default_pre_checks+=["some","new","checks"]
     pre_checks_url = {}
+    events = {}
+    commit_cache = {}
+    all_commits = []
+    ok_too_many_commits = False
+    warned_too_many_commits = False
+
+    l2s = ", ".join([gh_user_char + name for name in CMSSW_ISSUES_TRACKERS])
```
smuzaffar, line 747: I think this is not needed, you can move `", ".join([gh_user_char + name for name in CMSSW_ISSUES_TRACKERS])` to directly the place where we create the comment

process_pr.py
iarspider: Done

process_pr.py
iarspider: Done.

iarspider:
I have tested my changes with the following commands:
```
python3 process-pull-request -n 41162
python3 process-pull-request -n 11963
python3 process-pull-request -r cms-sw/cmsdist -n 8388
python3 process-pull-request -n 43064
python3 process-pull-request -n 43130
python3 process-pull-request -n 33532
```

process_pr.py
```diff
@@ -1003,6 +1013,21 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
+            technical_comment = comment
+
+        if (commenter == cmsbuild_user) and "This PR contains too many commits" in first_line:
+            warned_too_many_commits = True
+
+        if commenter in CMSSW_ISSUES_TRACKERS and re.match(
+            r"^\s*" + REGEX_IGNORE_COMMIT_COUNT + r"\s*$", first_line
+        ):
+            ok_too_many_commits = True
+
```
smuzaffar, line 1030: @iarspider , only one of these above conditions will match  so better to `continue`  instead of processing all

process_pr.py
```diff
@@ -666,6 +660,10 @@ def get_status_state(context, statuses):
     return ""
 
 
+def dumps_compact(value):
+    return dumps(value, separators=(",", ":"))
```
smuzaffar, line 664: @iarspider , better to sort keys to make sure we always get a unique output

process_pr.py
```diff
@@ -1003,6 +1013,22 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
                     pull_request_updated = True
             continue
 
+        if (
+            (commenter == cmsbuild_user)
+            and re.match(CMSBOT_TECHNICAL_MSG, first_line)
+            and not technical_comment
+        ):
+            technical_comment = comment
+
```
smuzaffar, line 1022: add `continue` here too

process_pr.py
smuzaffar: not really, it should be added in the `for` loop under the `if commit.sha not in commit_cache:` condition

process_pr.py
```diff
@@ -1254,38 +1348,120 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
+            issue.create_comment(new_body)
+    else:
+        if cache_comment:
+            print("DRY RUN: Updating existing comment with text")
+            print(new_body)
+        else:
+            print("DRY RUN: Creating technical comment with text")
+            print(new_body)
+
+    events = dict(sorted(events.items()))
```
smuzaffar, line 1421: we should not depend on the `dict` ordering here. As mentioned [here](https://docs.python.org/3.6/whatsnew/3.6.html#new-dict-implementation)
```
The order-preserving aspect of this new implementation is considered an implementation detail and should not be relied upon (this may change in the future, but it is desired to have this new dict implementation in the language for a few releases before changing the language spec to mandate order-preserving semantics for all current and future Python implementations; this also helps preserve backwards-compatibility with older versions of the language where random iteration order is still in effect, e.g. Python 3.5).
```

I would suggest to remove this and use something like `sorted(list(events.keys()))` directly in the `for` loop at line 1427

iarspider:
TODO: reset manually assigned categories on commit

process_pr.py
makortel: On the other hand https://docs.python.org/3/library/stdtypes.html#dict says
> Changed in version 3.7: Dictionary order is guaranteed to be insertion order. This behavior was an implementation detail of CPython from 3.6.

(I checked because I think we are already relying on the ordering behavior in several places in CMSSW)

(but maybe the sorted list of tuples is better anyway)

process_pr.py
```diff
@@ -1254,38 +1362,128 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
+    if cache_comment:
+        seen_commits_match = REGEX_COMMITS_CACHE.search(cache_comment.body)
+        if seen_commits_match:
+            print("Loading commit cache")
+            commit_cache = loads_maybe_decompress(seen_commits_match[1])
+
+    if pr.commits < MAX_INITIAL_COMMITS_IN_PR or ok_too_many_commits:
+        for commit in all_commits:
+            if commit.sha not in commit_cache:
+                api_rate_limits(gh)
```
smuzaffar, line 1402: remove this and add `check_rate_limits(msg=False)` in the `github_api()` itself

process_pr.py
```diff
@@ -1254,38 +1362,128 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
+            issue.create_comment(
+                f"This PR contains too many commits ({pr.commits} > {MAX_INITIAL_COMMITS_IN_PR}). "
+                "Make sure you chose the right target branch.\n"
+                f"{', '.join([gh_user_char + name for name in CMSSW_ISSUES_TRACKERS])}, "
+                "you can override this check with `+commit-count`."
+            )
+        return
 
-    # end of parsing comments section
+    api_rate_limits(gh)
```
smuzaffar, line 1381: remove this too

iarspider:
@smuzaffar ready for review.

process_pr.py
```diff
@@ -1149,9 +1217,25 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
                 signatures["code-checks"] = "rejected"
                 pre_checks_url["code-checks"] = comment.html_url
+                events[comment.created_at] = {
+                    "type": "sign",
+                    "value": {
+                        "ctype": "-1",
+                        "selected_cats": ["code-checks"],
+                        "comment": comment,
+                    },
+                }
```
smuzaffar, line 1227: @iarspider , I think we can remove this block

process_pr.py
```diff
@@ -1149,9 +1217,25 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
                 signatures["code-checks"] = "approved"
                 pre_checks_url["code-checks"] = comment.html_url
+                events[comment.created_at] = {
+                    "type": "sign",
+                    "value": {
+                        "ctype": "+1",
+                        "selected_cats": ["code-checks"],
+                        "comment": comment,
+                    },
+                }
```
smuzaffar, line 1238: @iarspider , I think we can remove this block

process_pr.py
```diff
@@ -1193,9 +1277,25 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
                     )
                     pre_checks_url["tests"] = comment.html_url
+                    events[comment.created_at] = {
+                        "type": "sign",
+                        "value": {
+                            "ctype": "+1",
+                            "selected_cats": ["tests"],
+                            "comment": comment,
+                        },
+                    }
```
smuzaffar, line 1287: @iarspider , I think we can remove this block

process_pr.py
```diff
@@ -1193,9 +1277,25 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
                     signatures["tests"] = "rejected"
                     pre_checks_url["tests"] = comment.html_url
+                    events[comment.created_at] = {
+                        "type": "sign",
+                        "value": {
+                            "ctype": "-1",
+                            "selected_cats": ["tests"],
+                            "comment": comment,
+                        },
+                    }
```
smuzaffar, line 1298: @iarspider , I think we can remove this block

process_pr.py
```diff
@@ -1212,6 +1312,14 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
                 if ("orp" in commenter_categories) and ("orp" in signatures):
                     signatures["orp"] = "approved"
+                    events[comment.created_at] = {
+                        "type": "sign",
+                        "value": {
+                            "ctype": "+1",
+                            "selected_cats": ["orp"],
+                            "comment": comment,
+                        },
+                    }
```
smuzaffar, line 1322: @iarspider , I think we can remove this block

process_pr.py
```diff
@@ -1242,50 +1350,146 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
+                elif ctype == "-1":
+                    for sign in selected_cats:
+                        signatures[sign] = "rejected"
+                        if sign == "orp":
+                            mustClose = False
+            else:
+                print(f"Ignoring event: {selected_cats} includes none of {signing_categories}")
+        elif event["type"] == "commit":
+            if cmssw_repo:
+                chg_categories = {"orp", "code-checks"}
```
smuzaffar, line 1474: I think we can set it to empty dict '{}' and keep the original signing status of `orp`, 'code-checks` and `tests`

process_pr.py
```diff
@@ -1242,50 +1320,134 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
+                        signatures[sign] = "rejected"
+            else:
+                print(f"Ignoring event: {selected_cats} includes none of {signing_categories}")
+        elif event["type"] == "commit":
+            if cmssw_repo:
+                # chg_categories = {"orp", "code-checks"}
+                # for fn in event["value"]:
+                #     chg_categories.update(
+                #         get_package_categories(cmssw_file2Package(repo_config, fn))
+                #     )
```
smuzaffar, line 1441: @iarspider , shouldn't we reset the signature for the cats touched by this commit?

process_pr.py
```diff
@@ -1434,15 +1434,13 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
-                # for fn in event["value"]:
-                #     chg_categories.update(
-                #         get_package_categories(cmssw_file2Package(repo_config, fn))
-                #     )
-                #
-                # chg_categories.update(assign_cats.keys())
-
-                for cat in assign_cats.keys():
+                for fn in event["value"]:
+                    chg_categories.update(
```
smuzaffar, line 1438: @iarspider ,  `chg_categories` needs to be defined before use ... right?

process_pr.py
```diff
@@ -1115,6 +1159,40 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
                     code_check_apply_patch = True
 
+        selected_cats = []
+
+        if re.match("^([+]1|approve[d]?|sign|signed)$", first_line, re.I):
+            ctype = "+1"
+            selected_cats = copy.copy(commenter_categories)
+        elif re.match("^([-]1|reject|rejected)$", first_line, re.I):
+            ctype = "-1"
+            selected_cats = copy.copy(commenter_categories)
```
smuzaffar, line 1169: @iarspider , why not just use `commenter_categories[:]` instead of `copy` module. `commenter_categories` holds just couple of strings

process_pr.py
```diff
@@ -1115,6 +1159,40 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
                 if m.group(2):
                     code_check_apply_patch = True
 
+        selected_cats = []
+
+        if re.match("^([+]1|approve[d]?|sign|signed)$", first_line, re.I):
+            ctype = "+1"
+            selected_cats = copy.copy(commenter_categories)
```
smuzaffar, line 1166: same here, `commenter_categories[:]` should be enough

process_pr.py
iarspider: Yes.

process_pr.py
iarspider: These are essentially the same.

smuzaffar: just change it to `commenter_categories[:]` please

smuzaffar:
I am getting this error 
```
> ./process-pull-request -n -r cms-sw/cmssw 11963
....
....
  File "./process-pull-request", line 83, in <module>
    process_pr(repo_config, gh, repo, repo.get_issue(prId), opts.dryRun, force=opts.force)
  File "/build/muz/bot/cms-bot/process_pr.py", line 1444, in process_pr
    get_package_categories(cmssw_file2Package(repo_config, fn))
ValueError: dictionary update sequence element #0 has length 14; 2 is required

```

smuzaffar:
can you please run bit more tests
```
> ./process-pull-request -n -r cms-sw/cmssw 43543
DRY RUN: Updating existing comment with text
  File "./process-pull-request", line 83, in <module>
    process_pr(repo_config, gh, repo, repo.get_issue(prId), opts.dryRun, force=opts.force)
  File "/build/muz/bot/cms-bot/process_pr.py", line 1406, in process_pr
    print(new_body)
UnicodeEncodeError: 'ascii' codec can't encode character '\u0144' in position 62: ordinal not in range(128)
````

process_pr.py
iarspider: done

smuzaffar:
+externals
This looks good now. I have tested it for various existsing PRs and https://github.com/cms-sw/cmssw/pull/43552. I will merge it tomorrow morning

smuzaffar:
There are too many commits so I will `squash and merge` this PR

PR #2115, opened 2023-11-16
Push GitHub API Rate Limits to ES

aandvalenzuela:
API remaining calls, limit, and reset time are pushed to the new index `cmssdt-github-api-*` on OpenSearch
Running on Jenkins under https://cmssdt.cern.ch/jenkins/view/All/job/github-monitor-api

### DIFF
--- a/github-rate-limits.py
+++ b/github-rate-limits.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 from __future__ import print_function
 import json, os
 from datetime import datetime
--- a/github-rate-limits.py
+++ b/github-rate-limits.py
@@ -18,31 +18,18 @@ if __name__ == "__main__":
     print("Max calls: ", limit)
     reset_time = datetime.fromtimestamp(gh.rate_limiting_resettime)
     print("Reset time (GMT): ", reset_time)
-
-    JENKINS_PREFIX = "jenkins"
-    try:
-        JENKINS_PREFIX = os.environ["JENKINS_URL"].strip("/").split("/")[-1]
-    except:
-        JENKINS_PREFIX = "jenkins"
+    gh_user = gh.get_user("cms-sw")
 
     current_time = datetime.utcnow() - datetime(1970, 1, 1)
     current_time = round(current_time.total_seconds() * 1000)
 
     gh_api_index = "cmssdt-github-api-" + str(int(((current_time / 86400000) + 4) / 7))
     gh_api_document = "github-api-data"
-    unique_id = (
-        JENKINS_PREFIX
-        + "/"
-        + str(reset_time).split(" ")[0].replace("-", "")
-        + "/"
-        + str(reset_time).split(" ")[1].replace(":", "")
-        + "/"
-        + str(remaining)
-    )
+    unique_id = str(gh_user) + str(current_time)
     unique_id = sha1(unique_id.encode()).hexdigest()
 
     payload = dict()
-    payload["jenkins_server"] = JENKINS_PREFIX
+    payload["jenkins_server"] = gh_user
     payload["api_limit"] = limit
     payload["api_remaining"] = remaining
     payload["reset_time"] = str(reset_time)
--- a/github-rate-limits.py
+++ b/github-rate-limits.py
@@ -18,18 +18,18 @@ if __name__ == "__main__":
     print("Max calls: ", limit)
     reset_time = datetime.fromtimestamp(gh.rate_limiting_resettime)
     print("Reset time (GMT): ", reset_time)
-    gh_user = gh.get_user("cms-sw")
+    user = os.getlogin()
 
     current_time = datetime.utcnow() - datetime(1970, 1, 1)
     current_time = round(current_time.total_seconds() * 1000)
 
     gh_api_index = "cmssdt-github-api-" + str(int(((current_time / 86400000) + 4) / 7))
     gh_api_document = "github-api-data"
-    unique_id = str(gh_user) + str(current_time)
+    unique_id = user + str(current_time)
     unique_id = sha1(unique_id.encode()).hexdigest()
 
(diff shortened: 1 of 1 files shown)


github-rate-limits.py
```diff
@@ -1,14 +1,53 @@
 #!/usr/bin/env python
 from __future__ import print_function
-from github import Github
```
smuzaffar, line 3: @aandvalenzuela , I think we can move to python3

github-rate-limits.py
```diff
@@ -1,14 +1,53 @@
...
+    print("Max calls: ", limit)
+    reset_time = datetime.fromtimestamp(gh.rate_limiting_resettime)
+    print("Reset time (GMT): ", reset_time)
+
+    JENKINS_PREFIX = "jenkins"
+    try:
+        JENKINS_PREFIX = os.environ["JENKINS_URL"].strip("/").split("/")[-1]
+    except:
+        JENKINS_PREFIX = "jenkins"
+
```
smuzaffar, line 27: remove `JENKINS_PREFIX` , it is not needed for monitoring the gh api rate

github-rate-limits.py
```diff
@@ -1,14 +1,53 @@
...
+    current_time = datetime.utcnow() - datetime(1970, 1, 1)
+    current_time = round(current_time.total_seconds() * 1000)
+
+    gh_api_index = "cmssdt-github-api-" + str(int(((current_time / 86400000) + 4) / 7))
```
smuzaffar, line 31: `str(int(((current_time / 86400000) + 4) / 7))` is week number ... right?

aandvalenzuela: Yes!

github-rate-limits.py
smuzaffar: I do not think `JENKINS_PREFIX` has any value for monitoring the api rate. Instead you should use github user name. So that if in future we have to monitor api rate for another user then we can make use of that.

github-rate-limits.py
```diff
@@ -1,14 +1,53 @@
...
+    gh_api_document = "github-api-data"
+    unique_id = (
+        JENKINS_PREFIX
+        + "/"
+        + str(reset_time).split(" ")[0].replace("-", "")
+        + "/"
+        + str(reset_time).split(" ")[1].replace(":", "")
+        + "/"
+        + str(remaining)
+    )
```
smuzaffar, line 41: just use github `"username"+ str(current_time)` for unique_id.

github-rate-limits.py
aandvalenzuela: It makes sense, thanks!

aandvalenzuela:
Find dashboard at https://monit-grafana.cern.ch/d/83t0M504k/cmssdt-jenkins?orgId=11&from=now-6h&to=now

github-rate-limits.py
```diff
@@ -18,31 +18,18 @@
     print("Max calls: ", limit)
     reset_time = datetime.fromtimestamp(gh.rate_limiting_resettime)
     print("Reset time (GMT): ", reset_time)
-
-    JENKINS_PREFIX = "jenkins"
-    try:
-        JENKINS_PREFIX = os.environ["JENKINS_URL"].strip("/").split("/")[-1]
-    except:
-        JENKINS_PREFIX = "jenkins"
+    gh_user = gh.get_user("cms-sw")
```
smuzaffar, line 21: this is one api call, is there any other way to get user with making an API call?

github-rate-limits.py
```diff
@@ -18,31 +18,18 @@
...
-        + str(reset_time).split(" ")[1].replace(":", "")
-        + "/"
-        + str(remaining)
-    )
+    unique_id = str(gh_user) + str(current_time)
     unique_id = sha1(unique_id.encode()).hexdigest()
 
     payload = dict()
-    payload["jenkins_server"] = JENKINS_PREFIX
+    payload["jenkins_server"] = gh_user
```
smuzaffar: why jenkins_server? why not just `user`?

smuzaffar: is there something special about `jenkins_server` field which I am not aware of :-)

aandvalenzuela: haha sorry, I missed that!

github-rate-limits.py
aandvalenzuela: Mmm... I don't see any property for getting the user. There is only one more apart from the `rate_limiting` and `rate_limiting_resettime`, but it doesn't return any user info ([oauth_scopes](https://pygithub.readthedocs.io/en/latest/github.html#github.MainClass.Github.oauth_scopes)):
```
['admin:org', 'gist', 'repo', 'user', 'workflow']
```

aandvalenzuela: Using the local linux user name instead :)

smuzaffar:
looks good to me, have you already tested it with latest changes?

aandvalenzuela:
Yes

smuzaffar:
humm, why I do not see the data via CERN Opensearch?

aandvalenzuela:
I see it [here](https://es-cmssdt.cern.ch/dashboards/app/data-explorer/discover#?_a=(discover:(columns:!(_source),isDirty:!f,sort:!()),metadata:(indexPattern:'3fc29390-88a1-11ea-8096-bf2a4db063ef',view:discover))&_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-20h,to:now))&_q=(filters:!(),query:(language:lucene,query:'_index:cmssdt-github*'))). There should be data since yesterday at 15h

smuzaffar:
no idea why I can not discover it by going to opensearch interface and then search for `api_limit` field. Anyway, graphana plot looks good. I am merging it now.

aandvalenzuela:
Mmm... I am looking into it, I also saw some errors on the shards yesterday that I would like to clarify

PR #2117, opened 2023-11-17
added results links for dqm diff and line changes

smuzaffar:


makortel:
I like the links, thanks!

PR #2120, opened 2023-11-29
add root label

smuzaffar:
bot should add label `root`  if `type root` comment is added by orp, l2's


smuzaffar:
FYI @Dr15Jones @makortel

PR #2123, opened 2023-12-04
Added l1scouting category

Mmiglio:
Added a new `l1scouting` category, introduced in the PR cms-sw/cmssw#43467. 
I'm not sure whether I can proceed with this PR or if I should wait for a formal request on the other PR.
Tagging @emeschi and @smorovic

### DIFF
--- a/categories.py
+++ b/categories.py
@@ -68,7 +68,7 @@ CMSSW_L2 = {
     "civanch": ["simulation", "geometry", "fastsim"],
     "bsunanda": ["geometry"],
     "davidlange6": ["operations"],
-    "emeschi": ["daq"],
+    "emeschi": ["daq", "l1scouting"],
     "vlimant": ["xpog"],
     "simonepigazzini": ["xpog"],
     "fwyzard": ["heterogeneous"],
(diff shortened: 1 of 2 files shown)


smuzaffar:
@Mmiglio , is there no existing category (e.g. `daq` or `l1`) who should be responsible for these new packages?

makortel:
> is there no existing category (e.g. `daq` or `l1`) which should be responsible for these new packages?

Let me tag @cms-sw/daq-l2 and @cms-sw/l1-l2 too. I think we'd need clear motivation to add a new signatory category.

aloeliger:
@Mmiglio Are there dedicated offline software conveners for l1Scouting? Is there an L1 scouting group operating in parallel to L1 that I am unaware of, or is this project managed under L1?

emeschi:
@aloeliger indeed the l1 scouting is part of the l1 project. I think it is perfectly fine if you and Enrique are willing to take care of the approval chain. It was me who actually suggested creating a new category since the l1 scouting, while being part of the trigger, does not participate in the trigger chain and is not acquired by the central daq, but has its own separate acquisition chain. For this reason, it may be convenient to keep it in a separate category, but I don't think anybody feels strongly about that. We can, if you wish, discuss this directly (in person or otherwise). @makortel we could indeed also put the code under the daq category, which would probably be more appropriate than l1 for the reasons stated above. Anyone of the three is okay with us.

emeschi:
Hello! Any news on this one ? As mentioned previously, we would be happy with any solution as long as this allows us to proceed with the actual PR for the code

smuzaffar:
I am fine with adding these packages under `daq`. If no objections for `l1` (@cms-sw/l1-l2 ) then @Mmiglio please update the PR

aloeliger:
No objections over here for this to go in DAQ.

PR #2133, opened 2023-12-13
github_utils: don't use f-strings

iarspider:


iarspider:
@smuzaffar This shoud fix syntax error here: https://cmssdt.cern.ch/jenkins/job/ib-run-pr-tests/36453, but the reason merge fails needs to be investigated

smuzaffar:
@iarspider , for now I have pushed https://github.com/cms-sw/cms-bot/commit/78616cb65067d3e5e4c5a959edfe70b13e9b9b0e to call new `merge_dicts` only for get_commit

smuzaffar:
these changes are now directly pushed

smuzaffar:
`github_utils.py` need to support both python2 and python3 ( as it is used by all cmssw release cycles). So changes like `f"string"` are not going to work in python2

PR #2136, opened 2023-12-15
allow to add non-blocking labels for non-cmssw repos

smuzaffar:
This change is to bot to support non-blocking labels for non-cmssw repos based on the change set

FYI @aandvalenzuela


smuzaffar:
please test

PR #2135, opened 2023-12-14
Use cms-docker issues to schedule and trigger builds for el8/el9 base image changes 

aandvalenzuela:
Used in https://cmssdt.cern.ch/jenkins/job/check-docker-container/

PR #2138, opened 2023-12-21
Report back to cms-docker issues when an image build has finished

aandvalenzuela:
Adding an option to just append a message to an existing issue (based on label match)


aandvalenzuela:
Now black is failing in a file that has been not touched `/home/runner/work/cms-bot/cms-bot/process_pr.py`.
FYI @smuzaffar, this PR only modifies the script I added in https://github.com/cms-sw/cms-bot/pull/2135 and it is not used by any other job

PR #2140, opened 2024-01-08
docker_launcher: create a separate voms proxy file per job ...

iarspider:
to avoid overwriting it - prevents errors like
```
DAS error: b'2024/01/08 01:25:37 ERROR failed to parse X509 proxy: crypto/tls: failed to parse key: asn1: syntax error: data truncated\n'
```

### DIFF
--- a/docker_launcher.sh
+++ b/docker_launcher.sh
@@ -20,6 +20,7 @@ else
 fi
 export DBS_URL=https://cmsweb.cern.ch:8443/dbs/prod/global/DBSReader
 export GIT_CONFIG_NOSYSTEM=1
+if [ "X$WORKSPACE" = "X" ] ; then export WORKSPACE=$(/bin/pwd) ; fi
 export X509_USER_PROXY=$WORKSPACE/x509up_u`id -u`
 XPATH=""
 py3or2_dir="$HOME/bin"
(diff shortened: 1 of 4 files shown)


smuzaffar:
@iarspider , thanks for looking in to this.
Instead of using 
```
-out $WORKSPACE/x509up_u`id -u`
```
why not just set 
```
export X509_USER_PROXY=$WORKSPACE/x509up_u`id -u`
```

and then all `voms-proxy-*`  will use the file pointing to this

iarspider:
@smuzaffar this is how it's done [here](https://github.com/cms-sw/cms-bot/blob/master/DMWM/setup-secrets.sh#L21). If setting `X509_USER_PROXY` in advance works, then I'll do it.

smuzaffar:
yes setting `X509_USER_PROXY` works ( at least  voms-proxy-init uses it to generate dump the new proxy). testing it by setting `X509_USER_PROXY` then using `voms-proxy-init` and `voms-proxy-info`

iarspider:
> yes setting `X509_USER_PROXY` works ( at least voms-proxy-init uses it to generate dump the new proxy). testing it by setting `X509_USER_PROXY` then using `voms-proxy-init` and `voms-proxy-info`

It doesn't work for `voms-proxy-destroy`, but we don't use it anywhere.

smuzaffar:
can you also move https://github.com/cms-sw/cms-bot/blob/d78dc1ac9e17e99966709dbffcfcb41985a05ba5/docker_launcher.sh#L52  above the `export X509_USER_PROXY...`

smuzaffar:
+externals

smuzaffar:
> X509_USER_PROXY doesn't work for `voms-proxy-destroy`, but we don't use it anywhere.

looks like a known issue https://github.com/italiangrid/voms-clients/issues/26 . Lets hope they include the fix https://github.com/italiangrid/voms-clients/pull/27 in next release

PR #2141, opened 2024-01-08
Add a parameter to disable comments on existing GH issues

aandvalenzuela:
The default behavior will be the same. By setting `-q True`, we can avoid commenting on the existing issue.

### DIFF
--- a/create-gh-issue.py
+++ b/create-gh-issue.py
@@ -10,14 +10,14 @@ SCRIPT_DIR = dirname(abspath(sys.argv[0]))
 
 parser = ArgumentParser()
 parser.add_argument(
-    "-r", "--repository", dest="repo", help="Github Repositoy name e.g cms-sw/cms-bot", type=str
+    "-r", "--repository", dest="repo", help="GitHub Repository name e.g cms-sw/cms-bot", type=str
 )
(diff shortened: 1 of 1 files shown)


create-gh-issue.py
```diff
@@ -66,8 +73,10 @@
     except:
         pass
 if issue:
-    print("Updating comment")
-    issue.create_comment(msg)
+    print(args.quiet)
+    if args.quiet == False:
+        print("Updating comment")
+        issue.create_comment(msg)
```
smuzaffar, line 79: do we really need an extra arg here? why not just return if issue already exist?

smuzaffar: Do we use this script `create-gh-issue.py` to add comments too?

smuzaffar: ok, looks like `run-user-pr-test` script is using it to either create issue or add a comment

aandvalenzuela: Since adding comments was the default behavior, I did not want to touch it. I see it is used in `run-user-pr-test` here:
* https://github.com/cms-sw/cms-bot/blob/master/run-user-pr-test#L37
* https://github.com/cms-sw/cms-bot/blob/master/run-user-pr-test#L88

Let me check if we really use the comment feature.

PR #2144, opened 2024-01-11
Detect commit squashing and preserve signatures

iarspider:


iarspider:
The function at https://github.com/cms-sw/cms-bot/pull/2144/files#diff-f37d46a488510bd0b96fd201ff189f8a704b756502df6a1cd46b117986e608a4R380 is needed because of API changes in python module. 

I would also prefer to keep commit cache in `bot_cache["commit"]`, similar to how reactions are kept in `bot_cache["emoji"]`. This way we won't need to filter `bot_cache` every time.

iarspider:
Tested here: https://github.com/iarspider-cmssw/cmssw/pull/10 . First squash was pure squash (i.e. no additional changes), second one was dirty (additional modification of README file)

process_pr.py
```diff
@@ -374,14 +377,25 @@ def modify_comment(comment, match, replace, dryRun):
...
+
 def set_comment_emoji_cache(dryRun, bot_cache, comment, repository, emoji="+1", reset_other=True):
     if dryRun:
         return
     comment_id = str(comment.id)
     if (
         (not comment_id in bot_cache["emoji"])
         or (bot_cache["emoji"][comment_id] != emoji)
-        or (comment.reactions[emoji] == 0)
+        or (get_comment_emojis_(comment)[emoji] == 0)
```
smuzaffar, line 398: why this was needed? I think `comment.reactions` should work for PyGithub deployed under /cvmfs/cms-ib.cern.ch  and cmssdt.

smuzaffar: https://github.com/PyGithub/PyGithub/pull/2813 merged upstream supports `comment.reactions` too

iarspider: My local version (which is probably newer than the one on cvmfs) doesn't have this property, instead it has `get_reactions` method returning list of reactions. I can change my local setup to use version from cvmfs and remove this wrapper.

iarspider: Latest release (2.1.1) [doesn't have](https://github.com/PyGithub/PyGithub/blob/v2.1.1/github/IssueComment.py#L124) `reactions` as property, but master [has](https://github.com/PyGithub/PyGithub/blob/main/github/IssueComment.py#L119). I will use version from cvmfs now.

process_pr.py
```diff
@@ -1,3 +1,5 @@
+from collections import defaultdict
```
smuzaffar, line 1: is this needed any more?

process_pr.py
```diff
@@ -871,6 +874,7 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
                 if pr.base.ref == "master":
                     signing_categories.add("code-checks")
                 updateMilestone(repo, issue, pr, dryRun)
+                pass
```
smuzaffar, line 877: ??

process_pr.py
```diff
@@ -1459,7 +1460,50 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
+                            x["filename"]
+                            for x in get_commit(repo.full_name, new_head_commit_sha)["files"]
+                        ),
+                    }
+
+                # Clean removed commits from cache
+                for k in commit_cache:
+                    if not any(k == commit.sha for commit in all_commits):
+                        print("Removing commit {0} from cache".format(k))
+                        del bot_cache[k]
```
smuzaffar, line 1505: @iarspider , should we move this block outside the `if all_commits.totalCount < seen_commits_cnt:` condition? e.g. for this PR itself the bot cache has 8 commits while there are only 3 commits ( I think due to force pushes). And I think this block can go in a separate PR as it fixes the issue of invalid commits in bot cache

iarspider: I'm not sure if we will gain anything from moving this block outside of `if` condition - it will just waste CPU cycles for PRs where that condition is false (i.e. any normal PR, without force-pushes). Regarding putting this in a separate PR - well, we can, but that new PR should go in before this one (maybe also use that new PR to move commits under their own key in cache)

smuzaffar: yes, you are right this block will not do any thing for most PRs but it is useful to clean the commits for those PRs where original commits (before force push) and new commit count is same. e.g. PR with single commit with N force pushes will have N+1 commits in the bot cache.

Ok lets leave the block in this PR

process_pr.py
```diff
@@ -1459,7 +1460,50 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
+
+                if diff_before_squash ^ diff_after_squash:
+                    print("PR diff changed, will not preserve signatures")
+                else:
+                    print("PR diff not changed, preserving signatures")
+                    bot_cache[new_head_commit_sha] = {
+                        "time": bot_cache[last_seen_commit_sha]["time"] + 1,
+                        "files": sorted(
+                            x["filename"]
+                            for x in get_commit(repo.full_name, new_head_commit_sha)["files"]
```
smuzaffar, line 1497: @iarspider , this does not look right e.g. lets say we have

- commit1 at T0: touching core and reco files
- core signs PR at T1
- commit2 at T2: touching reco files
- bot see commit2 and it is now our last_seen_commit
- squash_commit at T2+1 will have all files core and reco which means core signatures will not be preserved ... right?

iarspider:
Tested on https://github.com/iarspider-cmssw/cmssw/pull/11

process_pr.py
```diff
@@ -1496,6 +1589,17 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
                     }
                 )
 
+            # Inject events from cached commits
+            for k, cache_entry in commit_cache.items():
```
smuzaffar, line 1593: @iarspider , should this be done for only `squashed` commits?

process_pr.py
```diff
@@ -1459,35 +1464,123 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
+                diff_after_squash = {
+                    (file.filename, file.patch) for file in changes_after_squash.files
+                }
+
+                if diff_before_squash ^ diff_after_squash:
+                    print("PR diff changed, will not preserve signatures")
+                else:
+                    print("PR diff not changed, preserving signatures and commit statuses")
+                    bot_cache[new_head_commit_sha] = {
+                        "time": bot_cache[last_seen_commit_sha]["time"] + 1,
```
smuzaffar, line 1517: @iarspider , any specific reasons to not use the `time` as it is (instead of adding `1`)?

process_pr.py
```diff
@@ -1496,6 +1589,17 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
                     }
                 )
 
+            # Inject events from cached commits
+            for k, cache_entry in commit_cache.items():
+                if cache_entry["time"] not in events:
```
smuzaffar, line 1594: @iarspider , this might miss the commits with same timestamp

process_pr.py
```diff
@@ -1464,88 +1471,198 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
+        print("Event:", event)
+        if event["type"] == "sign":
+            comment = event["value"]["comment"]
+            cached_signed_comments = bot_cache["signatures"].get(comment.id)
+            if cached_signed_comments and cached_signed_comments != signed_commit_sha:
+                print(
+                    "WARNING: For comment {0}, cached list of signed commits doesn't match the present list of signed commits. All signatures will be reset."
+                )
+                reset_all_signatures = True
+                break
```
smuzaffar, line 1616: @iarspider , we should not stop processing the event loop here. We can set thumb's down (so some other) emoji for this comment and ignore this comment. We still want L2 to sign such PRs ... right? So if we break the loop then how can L2 sign it

process_pr.py
```diff
@@ -1,3 +1,5 @@
+import hashlib
```
smuzaffar, line 1: @iarspider , this is not needed any more ... right?

process_pr.py
```diff
@@ -1464,88 +1471,197 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
-                ctype = event["value"]["ctype"]
-                if any(x in signing_categories for x in selected_cats):
-                    set_comment_emoji_cache(
-                        dryRun, bot_cache, event["value"]["comment"], repository
+        flattened_eventlist.extend(eventlist)
+
+    for event in flattened_eventlist:
+        print("Event:", event)
+        if event["type"] == "sign":
+            comment = event["value"]["comment"]
```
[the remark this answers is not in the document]
iarspider: @smuzaffar not sure if it'll work as expected. You want to ignore signatures made before squashed commit, right? But right now `signed_commit_sha` will be updated for each commit, including commits that disappeared during squash. And if we check for `squashed` flag to filter out removed commits, I think this defeats the purpose of this PR - to make squash commits transparent.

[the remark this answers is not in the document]
smuzaffar: no , I want to ignore the sign if `signed_commit_sha` is not set. This could happen if someone force push a change set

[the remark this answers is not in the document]
smuzaffar: ah right, thinking about it again, you are right. In case of force push the sign in bot cache should have the previous commit hash

iarspider:
Not sure why code-style check fails.

process_pr.py
```diff
@@ -1466,86 +1470,194 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
+                        signatures[sign] = "rejected"
+            else:
+                print(
+                    "Ignoring event: {0} includes none of {1}".format(
+                        selected_cats, signing_categories
+                    )
+                )
+        elif event["type"] == "commit":
+            if event["value"]["files"]:
+                signed_commit_sha = event["value"]["sha"]
```
smuzaffar, line 1646: @iarspider , I think the condition `if event["value"]["files"]` is not needed. ` signed_commit_sha = event["value"]["sha"]` should work for all commits events

process_pr.py
```diff
@@ -1466,86 +1470,194 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
-                ctype = event["value"]["ctype"]
-                if any(x in signing_categories for x in selected_cats):
-                    set_comment_emoji_cache(
-                        dryRun, bot_cache, event["value"]["comment"], repository
+        flattened_eventlist.extend(eventlist)
+
+    for event in flattened_eventlist:
+        print("Event:", event)
+        if event["type"] == "sign":
+            comment = event["value"]["comment"]
```
smuzaffar, line 1609: @iarspider , what will happen if `signed_commit_sha` is not set ( e.g we have a `+1` before the commit time which can happen when one force push a change after some one has laready signed the previous commit). Do we need to all the logic for `approved/rejected`  in that case? Can we simply skip such comment? e.g. `if not signed_commit_sha: continue` here

smuzaffar: ah never mind, I forgot that the removed commit will still be in event loop

smuzaffar:
looks good, let me test it for https://github.com/cms-sw/cmssw/pull/43806

process_pr.py
```diff
@@ -1466,86 +1470,194 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
+            cached_signed_commit_sha = bot_cache["signatures"].get(comment.id)
```
smuzaffar, line 1610: @iarspider , this failed with error
```
  File "/tmp/cmsbuild/jenkins/workspace/cms-bot/cms-bot/process-pull-request.py", line 83, in <module>
    process_pr(repo_config, gh, repo, repo.get_issue(prId), opts.dryRun, force=opts.force)
  File "/tmp/cmsbuild/jenkins/workspace/cms-bot/cms-bot/process_pr.py", line 1610, in process_pr
    cached_signed_commit_sha = bot_cache["signatures"].get(comment.id)
KeyError: 'signatures'
```

process_pr.py
iarspider: Do we want to sign squash/merge commits? This check is there to skip such commits.

process_pr.py
iarspider: Fixed.

process_pr.py
```diff
@@ -1466,86 +1479,191 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
                 events[datetime.fromtimestamp(cache_entry["time"])].append(
-                    {
-                        "type": "commit",
-                        "value": cache_entry["files"],
-                    }
+                    {"type": "commit", "value": {"files": cache_entry["files"], "sha": commit.sha}}
                 )
 
+            # Inject events from cached commits
+            for commit_sha in missing_commits:
```
smuzaffar, line 1596: @iarspider , I think we need to process all commits from the cache

process_pr.py
```diff
@@ -1239,8 +1252,12 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
         elif re.match(r"^\s*" + REGEX_IGNORE_COMMIT_COUNT + r"\s*$", first_line):
             comment_emoji = "-1"
             if commenter in CMSSW_ISSUES_TRACKERS:
-                ok_too_many_commits = True
-                comment_emoji = "+1"
+                if pr.commits < TOO_MANY_COMMITS_FAIL_THRESHOLD:
+                    ok_too_many_commits = True
+                    comment_emoji = "+1"
+                else:
+                    warned_too_many_commits = False
```
smuzaffar, line 1259: @iarspider , do we need to reset this here? what will happen if this comment is after bot has warned about the too many commits?

process_pr.py
```diff
@@ -219,11 +222,24 @@ def init_l2_data(repo_config, cms_repo):
     return l2_data
 
 
+def collect_commit_cache(bot_cache):
+    commit_cache = {k: v for k, v in bot_cache.items() if REGEX_COMMIT_SHA.match(k)}
```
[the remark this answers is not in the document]
iarspider: Ok.

process_pr.py
```diff
@@ -219,11 +222,24 @@ def init_l2_data(repo_config, cms_repo):
...
+        bot_cache["commits"] = commit_cache
+
+
 def read_bot_cache(comment_msg):
     seen_commits_match = REGEX_COMMITS_CACHE.search(comment_msg)
     if seen_commits_match:
         print("Loading bot cache")
-        return loads_maybe_decompress(seen_commits_match[1])
+        res = loads_maybe_decompress(seen_commits_match[1])
+        for k, v in BOT_CACHE_TEMPLATE.items():
```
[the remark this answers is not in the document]
iarspider: I wanted to keep it global for visibility, but maybe moving it inside is also a good idea.

process_pr.py
iarspider: yes, it's a typo - I wanted to reset `ok_too_many_commits`

process_pr.py
```diff
@@ -1239,8 +1253,12 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
             comment_emoji = "-1"
             if commenter in CMSSW_ISSUES_TRACKERS:
-                ok_too_many_commits = True
-                comment_emoji = "+1"
+                if pr.commits < TOO_MANY_COMMITS_FAIL_THRESHOLD:
+                    ok_too_many_commits = True
+                    comment_emoji = "+1"
+                else:
+                    ok_too_many_commits = False
+                    comment_emoji = "-1"
```
smuzaffar, line 1261: so this block can be simplify to
```
ok_too_many_commits = pr.commits < TOO_MANY_COMMITS_FAIL_THRESHOLD
comment_emoji = "+1" if ok_too_many_commits else "-1"
```
right?

iarspider: Yes. We can also always set emoji to `+1` if we want to keep `-1` as indication of missing permission or wrong syntax.

process_pr.py
```diff
@@ -1460,92 +1480,217 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
+
+            elif len(commit.parents) > 1:
+                bot_cache["commits"][commit.sha]["files"] = []
+
+            if (
+                last_seen_commit_time
+                and commit.commit.committer.date.timestamp() > last_seen_commit_time
+            ):
+                bot_cache["commits"][commit.sha]["files"] = []
+                bot_cache["commits"][commit.sha]["time"] = last_seen_commit_time
```
smuzaffar, line 1615: @iarspider , are you still working on the part which we discussed about setting empty list of files for all commits which are in PR but not in cache if there are no changes between last seen and last pushed changes?

PR #2153, opened 2024-01-19
Including Geometry/HGCalMapping in categories_map

kerstinlovisa:
Including Geometry/HGCalMapping in categories_map.py under "geometry"

### DIFF
--- a/categories_map.py
+++ b/categories_map.py
@@ -815,6 +815,7 @@ CMSSW_CATEGORIES = {
         "Geometry/HGCalGeometry",
         "Geometry/HGCalSimData",
         "Geometry/HGCalTBCommonData",
+        "Geometry/HGCalMapping",
         "Geometry/HcalAlgo",
         "Geometry/HcalCommonData",
         "Geometry/HcalEventSetup",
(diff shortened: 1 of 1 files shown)


iarspider:
assign geometry

pfs:
@kerstinlovisa @all  i would suggest to add to `upgrades` as well for the moment

iarspider:
assign geometry, upgrades

kerstinlovisa:
@pfs @ALL HGCalMapping also added to upgrades category now

categories_map.py
```diff
@@ -1632,6 +1633,7 @@
     "upgrade": [
         "CalibTracker/SiPhase2TrackerESProducers",
         "CondFormats/HGCalObjects",
+        "Geometry/HGCalMapping",
```
makortel, line 1636: Please move this downwards to keep the alphabetic order.

kerstinlovisa: Moved it down so it should be in alphabetic order now

smuzaffar:
any objections @cms-sw/geometry-l2 @cms-sw/upgrade-l2  on adding `Geometry/HGCalMapping` under your control?

civanch:
+1

PR #2156, opened 2024-01-23
Accept progress labels for cms-docker issues

aandvalenzuela:


repos/cms_sw/cms_docker/githublabels.py
```diff
@@ -9,3 +9,12 @@
...
     TYPE_COMMANDS[arch] = [LABEL_COLORS["doc"], "%s-[0-9a-f]+" % arch, "mtype", True]
+    for status in [
+        ("build-queued", "pending"),
+        ("building", "pending"),
+        ("build-done", "approved"),
+        ("build-error", "rejected"),
+    ]:
+        TYPE_COMMANDS[arch].append(
+            [LABEL_COLORS[status[1]], "%s[-a-z]*%s" % (arch, status[0]), "mtype", True]
+        )
```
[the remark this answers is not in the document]
aandvalenzuela: Thanks @smuzaffar! Sounds like a good idea. For now, I was simply removing the labels of the previous state by using the `remove_issue_label` method from the [GH utils](https://github.com/cms-sw/cms-bot/blob/master/github_utils.py#L518). I think we do the same in the Build Release jobs.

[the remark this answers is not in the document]
smuzaffar: @aandvalenzuela , bot will keep adding the labels back due to `type ....` comments so removing a label will not help

repos/cms_sw/cms_docker/githublabels.py
```diff
@@ -9,16 +9,10 @@
...
-        arch_status = "%s-%s" % (arch, status[0])
-        TYPE_COMMANDS[arch_status] = [
-            LABEL_COLORS[status[1]],
-            "%s[-a-z]*%s" % (arch, status[0]),
-            "state",
-            True,
-        ]
+    TYPE_COMMANDS[arch + "-cms-docker"] = [
+        LABEL_COLORS["doc"],
+        "%s-[a-z]+-(queued|building|done|error)" % arch,
```
smuzaffar, line 14: @aandvalenzuela ,  this regexp should match `arch-something-(.*)` but what about `x86_64` tag itself

process_pr.py
```diff
@@ -483,6 +483,7 @@ def check_extra_labels(first_line, extra_labels):
 def check_type_labels(first_line, extra_labels):
     ex_labels = {}
     rem_labels = {}
+    state_labels = {}
```
smuzaffar, line 486: what is the purpose of this if it is not used any where :-) you can pass it as a input parameters to `check_type_labels` and then make use of it before https://github.com/cms-sw/cms-bot/blob/master/process_pr.py#L1602

process_pr.py
```diff
@@ -494,7 +495,9 @@ def check_type_labels(first_line, extra_labels):
                 obj_labels = rem_labels if rem_lab else ex_labels
                 if lab_type not in obj_labels:
                     obj_labels[lab_type] = []
-                if (len(TYPE_COMMANDS[lab]) > 3) and TYPE_COMMANDS[lab][3]:
+                if (len(TYPE_COMMANDS[lab]) > 4) and TYPE_COMMANDS[lab][4] == "state":
+                    state_labels[lab_type].append(type_cmd)
```
smuzaffar, line 499: it should be `state_labels[lab].append(lab_type)` where `lab`  will be `<arch>-cms-docker` and  `lab_type` will be the  type comemnt value e.g. `<arch>-runtime-building`

smuzaffar: @aandvalenzuela , I editted the cooment above . This should be `state_labels[lab]=lab_type`

repos/cms_sw/cms_docker/githublabels.py
smuzaffar: @aandvalenzuela , I would suggest to change this to `"%s-([a-z]+-|)(queued|building|done|error)" % arch`

repos/cms_sw/cms_docker/githublabels.py
```diff
@@ -11,7 +11,7 @@
     TYPE_COMMANDS[arch] = [LABEL_COLORS["doc"], "%s-[0-9a-f]+" % arch, "mtype", True]
     TYPE_COMMANDS[arch + "-cms-docker"] = [
         LABEL_COLORS["doc"],
-        "%s-[a-z]+-(queued|building|done|error)" % arch,
+        "%s[-a-z]*-(queued|building|done|error)" % arch,
```
smuzaffar, line 14: @aandvalenzuela , this is still not correct as it can match `x86_64foobar-queued` i.e. tag part starting with `[a-z]` instead of `-`

process_pr.py
```diff
@@ -1595,15 +1600,16 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
     print("Old Labels:", sorted(old_labels))
     print("Compilation Warnings: ", comp_warnings)
     print("Singnatures: ", signatures)
     if "mtype" in extra_labels:
         extra_labels["mtype"] = list(set(extra_labels["mtype"]))
     if "type" in extra_labels:
         extra_labels["type"] = [extra_labels["type"][-1]]
-
+    if len(state_labels) != 0:
+        extra_labels["mtype"].append(list(state_labels.values()))
```
smuzaffar, line 1612: @aandvalenzuela , this should be `extend` instead of `append` :-)

process_pr.py
```diff
@@ -1595,15 +1600,16 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
+    print("Stats:", backport_pr_num, extra_labels, state_labels)
     print("Old Labels:", sorted(old_labels))
     print("Compilation Warnings: ", comp_warnings)
     print("Singnatures: ", signatures)
     if "mtype" in extra_labels:
         extra_labels["mtype"] = list(set(extra_labels["mtype"]))
     if "type" in extra_labels:
         extra_labels["type"] = [extra_labels["type"][-1]]
-
+    if len(state_labels) != 0:
```
smuzaffar, line 1611: there is no need to add this check. `extra_labels["mtype"].extend(state_labels.values())` should work for empty `state_labels` too

smuzaffar:
+externals

done tests for https://github.com/cms-sw/cms-docker/issues/266 and bot process the new state types correctly

PR #2161, opened 2024-01-30
[cms-docker] Manage state labels and correct pulls curl

aandvalenzuela:


cms-docker-gh-issues.py
```diff
@@ -69,7 +69,8 @@
 )
 
 if args.comment == False:
-    pulls_curl = "curl -s 'https://api.github.com/repos/%s/pulls?q=+is:open+label:%s'" % (
+
+    pulls_curl = "curl -s 'https://api.github.com/repos/%s/issues?state=open&labels=%s'" % (
```
smuzaffar, line 73: by the way, why are you not using `https://api.github.com/repos/%s/issues` API call  for `issues_curl` ?

aandvalenzuela: For issues I was using the same call as in https://github.com/cms-sw/cms-bot/blob/master/create-gh-issue.py#L63, but I can modify it :)

cms-docker-gh-issues.py
```diff
@@ -91,20 +92,24 @@
...
 
-        # Get current issue number
-        print("Check newly created Issue", issues_curl)
-        exit_code, issues_obj = run_cmd(issues_curl)
-        issues_dict = json.loads(issues_obj)
-        issue_number = issues_dict["items"][0]["number"]
+        issues = gh_repo.get_issues(labels=[str(label) for label in args.labels])
+
+        for issue in issues:
+            issue_number = issue.number
```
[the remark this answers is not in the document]
aandvalenzuela: Right, I left it behind when debugging. We should have only one matching issue.

cms-docker-gh-issues.py
```diff
@@ -91,20 +92,22 @@
...
+                urls += "* " + str(pull_obj.get("html_url")) + "\n"
+        print("The following PRs have matching labels: \n", urls)
 
-        # Get current issue number
-        print("Check newly created Issue", issues_curl)
-        exit_code, issues_obj = run_cmd(issues_curl)
-        issues_dict = json.loads(issues_obj)
-        issue_number = issues_dict["items"][0]["number"]
+        issues = gh_repo.get_issues(labels=[str(label) for label in args.labels])
+        issue_number = issue.number
```
smuzaffar, line 101: @aandvalenzuela , this will fail as there is no `issue`. For this section why not use the issue returned by `gh_repo.create_issue` call?

aandvalenzuela:
Merging since it has been running fine for the last few days

PR #2163, opened 2024-01-31
[draft] Mark all new commits as having no files on clean squash

iarspider:
To be tested

### DIFF
--- a/process_pr.py
+++ b/process_pr.py
@@ -867,7 +867,7 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
     pre_checks_url = {}
     events = defaultdict(list)
     all_commits = []
-    all_commit_shas = []
+    all_commit_shas = set()
     ok_too_many_commits = False
     warned_too_many_commits = False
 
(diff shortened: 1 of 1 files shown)


process_pr.py
```diff
@@ -1517,7 +1517,7 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
         missing_commits = set(
             k for k in bot_cache["commits"] if not bot_cache["commits"][k].get("squashed", False)
         ).difference(all_commit_shas)
-        last_seen_commit_time = None
```
smuzaffar, line 1520: why remove this?

process_pr.py
```diff
@@ -1517,7 +1517,7 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
+        new_commits = all_commit_shas.difference(k for k in bot_cache["commits"])
```
smuzaffar, line 1520: why not move it under the condition when there are no differences between last seen and new commits i.e under https://github.com/cms-sw/cms-bot/pull/2144/files#diff-f37d46a488510bd0b96fd201ff189f8a704b756502df6a1cd46b117986e608a4R1550 ?

process_pr.py
```diff
@@ -1607,10 +1607,7 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
             elif len(commit.parents) > 1:
                 bot_cache["commits"][commit.sha]["files"] = []
 
-            if (
-                last_seen_commit_time
-                and commit.commit.committer.date.timestamp() > last_seen_commit_time
-            ):
+            if last_seen_commit_time is not None and commit.sha in new_commits:
```
smuzaffar, line 1610: once calculation of `new_commits` is moved at the right place then no need to check for `last_seen_commit_time is not None`  (just `commit.sha in new_commits` should be enough)

smuzaffar:
looks good

PR #2169, opened 2024-02-08
Automatically merge cmsdist PR that updates cmsswdata if correspondin…

iarspider:
…g cms-data PR was for the default branch

### DIFF
--- a/create-new-data-pr.py
+++ b/create-new-data-pr.py
@@ -203,3 +203,7 @@ if __name__ == "__main__":
     change_tag_pull_request = dist_repo.create_pull(
         title=title, body=body, base=default_cms_dist_branch, head=repo_tag_pr_branch
     )
+
+    if data_pr_base_branch == data_repo_default_branch:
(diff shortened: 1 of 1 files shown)


create-new-data-pr.py
```diff
@@ -203,3 +203,7 @@ def get_tag_from_string(tag_string=None):
     change_tag_pull_request = dist_repo.create_pull(
         title=title, body=body, base=default_cms_dist_branch, head=repo_tag_pr_branch
     )
+
+    if data_pr_base_branch == data_repo_default_branch:
+        print("cms-data PR was for default branch, will merge now")
```
smuzaffar, line 208: @iarspider , may be add a comment to PR that why it is beening automatically merged.
Also I would suggest to a add a new command-line option e.g. `--no-merge`  (default value `False`) so we can use it (if needed) to only create the PR without merge

smuzaffar:
+externals

tested via creating test PR [iarspider/cmsdist#1](https://github.com/iarspider/cmsdist/pull/1)

PR #2174, opened 2024-02-15
Always return list from get_pr_commits_reversed

iarspider:


smuzaffar:
@iarspider , looks like this has not fixed the issue of extra api call for commits.

PR #2177, opened 2024-02-16
Drop user name (user login is enough) from bot messages to avoid one api call

smuzaffar:
Every time bot runs (even if there are no changes for PR or issue) bot make an api call to get gihtub user name. This change proposes to use only github user login instead of user login + user name. So after this change bot will post message like 
```
A new Pull Request was created by @smuzaffar for branch master.
```
 instead of
```
A new Pull Request was created by @smuzaffar (Malik Shahzad Muzaffar) for branch master.
```

If we want to keep the user name then we can update bot to only generate the comment messages when needed e.g. if bot has already seen the issue/PR then no need to execute https://github.com/cms-sw/cms-bot/blob/master/process_pr.py#L2322-L2345 code.


smuzaffar:
+externals

PR #2179, opened 2024-02-21
Modify exit codes for cms-docker building issues

aandvalenzuela:
This PR modifies the exit codes that are then handled in the Jenkins job. With the new values, if there is any error when running the script, the Jenkins job will nicely clean up the corresponding property files.


aandvalenzuela:
Tested on Jenkins

PR #2183, opened 2024-02-23
Fix handling of big cache

iarspider:
@smuzaffar following discussion in https://github.com/cms-sw/cmssw/pull/44025. This will cause us a few extra API calls, and will likely make squash detection less efficient, but at least the tests will run.

### DIFF
--- a/process_pr.py
+++ b/process_pr.py
@@ -186,6 +186,7 @@ MULTILINE_COMMENTS_MAP = {
     + "|_input|)": [RELVAL_OPTS, "EXTRA_MATRIX_COMMAND_ARGS", True],
 }
 
+BOT_CACHE_CHUNK_SIZE = 55000
 TOO_MANY_COMMITS_WARN_THRESHOLD = 150
 TOO_MANY_COMMITS_FAIL_THRESHOLD = 240
 L2_DATA = {}
@@ -264,8 +265,9 @@ def prepare_bot_cache(bot_cache):
         # The limit is 65535 chars, and minimal bot cache comment is
         # `cms-bot internal usage<!-- bot cache:  -->`, 42 characters.
         # But since the cache can be embedded in an "already seen" comment, whose length is not
-        # known in advance, we limit the amout of (encoded cache) data per comment to 55k chars.
-        part, data = data[:55000], data[55000:]
+        # known in advance, we limit the amout of (encoded cache) data per comment
+        # to BOT_CACHE_CHUNK_SIZE chars.
+        part, data = data[:BOT_CACHE_CHUNK_SIZE], data[BOT_CACHE_CHUNK_SIZE:]
         res.append(part)
 
     return res
@@ -798,7 +800,7 @@ def get_status_state(context, statuses):
 
 def dumps_maybe_compress(value):
     json_ = dumps(value, separators=(",", ":"), sort_keys=True)
-    if len(json_) > 55000:
+    if len(json_) > BOT_CACHE_CHUNK_SIZE:
         return "b64:" + base64.encodebytes(zlib.compress(json_.encode())).decode("ascii", "ignore")
     else:
         return json_
--- a/process_pr.py
+++ b/process_pr.py
@@ -49,7 +49,6 @@ try:
 except ImportError:
     from yaml import Loader, Dumper
 
-
 try:
     from categories import CMSSW_LABELS
 except:
@@ -135,7 +134,7 @@ REGEX_TEST_IGNORE = re.compile(
     r"^\s*(?:(?:@|)cmsbuild\s*[,]*\s+|)(?:please\s*[,]*\s+|)ignore\s+tests-rejected\s+(?:with|)([a-z -]+)$",
     re.I,
 )
-REGEX_COMMITS_CACHE = re.compile(r"<!-- (?:commits|bot) cache: (.*) -->")
+REGEX_COMMITS_CACHE = re.compile(r"<!-- (?:commits|bot) cache: (.*) -->", re.DOTALL)
 REGEX_IGNORE_COMMIT_COUNT = "\+commit-count"
 TEST_WAIT_GAP = 720
 ALL_CHECK_FUNCTIONS = None
@@ -230,31 +229,62 @@ def collect_commit_cache(bot_cache):
         bot_cache["commits"] = commit_cache
 
 
-def read_bot_cache(comment_msg):
-    seen_commits_match = REGEX_COMMITS_CACHE.search(comment_msg)
-    if seen_commits_match:
-        print("Loading bot cache")
-        res = loads_maybe_decompress(seen_commits_match[1])
-        for k, v in BOT_CACHE_TEMPLATE.items():
-            if k not in res:
-                res[k] = v
-        collect_commit_cache(res)
-        return res
+def read_bot_cache(data):
+    print("Loading bot cache")
+    res = loads_maybe_decompress(data)
+    for k, v in BOT_CACHE_TEMPLATE.items():
+        if k not in res:
+            res[k] = v
+    collect_commit_cache(res)
+    return res
+
+
+def extract_bot_cache(comment_msgs):
+    print(
+        "Reading bot cache from technical comment(s):",
+        ",".join(str(c) for c in comment_msgs),
+    )
+
+    data = ""
+    for comment_msg in comment_msgs:
+        seen_commits_match = REGEX_COMMITS_CACHE.search(comment_msg.body)
+        if seen_commits_match:
+            data += seen_commits_match[1]
+
+    if data:
+        return read_bot_cache(data)
+
     return {}
 
 
-def write_bot_cache(bot_cache, cache_comment, issue, dryRun):
-    old_body = cache_comment.body if cache_comment else CMSBOT_TECHNICAL_MSG
-    new_body = (
-        REGEX_COMMITS_CACHE.sub("", old_body)
-        + "<!-- bot cache: "
-        + dumps_maybe_compress(bot_cache)
-        + " -->"
-    )
-    if old_body == new_body:
-        return
-    print("Saving bot cache")
-    if len(new_body) <= 65535:
+def prepare_bot_cache(bot_cache):
+    res = []
+    data = dumps_maybe_compress(bot_cache)
+    while data:
+        # The limit is 65535 chars, and minimal bot cache comment is
+        # `cms-bot internal usage<!-- bot cache:  -->`, 42 characters.
+        # But since the cache can be embedded in an "already seen" comment, whose length is not
+        # known in advance, we limit the amout of (encoded cache) data per comment to 55k chars.
+        part, data = data[:55000], data[55000:]
+        res.append(part)
+
+    return res
+
+
+def write_bot_cache(bot_cache, cache_comments, issue, dryRun):
+    data = prepare_bot_cache(bot_cache)
+    for i, part in enumerate(data):
+        try:
+            cache_comment = cache_comments[i]
+        except (TypeError, IndexError):
+            cache_comment = None
+
+        old_body = cache_comment.body if cache_comment else CMSBOT_TECHNICAL_MSG
+        new_body = REGEX_COMMITS_CACHE.sub("", old_body) + "<!-- bot cache: " + part + " -->"
+        if old_body == new_body:
+            continue
+
+        print("Saving bot cache ({0}/{1})".format(i + 1, len(data)))
         if not dryRun:
             if cache_comment:
                 cache_comment.edit(new_body)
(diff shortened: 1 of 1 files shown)


smuzaffar:
@iarspider , why not break the large comment in multiple and concatenate while reading?

iarspider:
I have considered it, but decided that just ignoring the overflow would be
good enough, considering how rare this issue is. I will prepare a new PR
then.

process_pr.py
```diff
@@ -1074,7 +1124,7 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
             )
         print("Pre check status:", pre_checks_state)
     already_seen = None
-    technical_comment = None
+    technical_comments = {}
```
smuzaffar, line 1127: @iarspider , having this as `list` will reduce some code

process_pr.py
```diff
@@ -266,8 +296,28 @@ def write_bot_cache(bot_cache, cache_comment, issue, dryRun):
             else:
                 print("DRY RUN: Creating technical comment with text")
             print(new_body.encode("ascii", "ignore").decode())
-    else:
-        raise RuntimeError("Updated comment body too long: {0} > 65535".format(len(new_body)))
+
+    # If new commit cache is smaller than previous one, cleanup old technical comments
+    if len(data) < len(cache_comments):
+        for i in range(len(data), len(cache_comments)):
+            cache_comment = cache_comments[i]
```
smuzaffar, line 303: @iarspider , I would suggest to just delete the comment

process_pr.py
```diff
@@ -1152,10 +1210,10 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
                     pull_request_updated = comment.created_at < last_commit_date
                 continue
-            if re.match(CMSBOT_TECHNICAL_MSG, first_line):
-                if not technical_comment:
-                    technical_comment = comment
-                continue
+            # if re.match(CMSBOT_TECHNICAL_MSG, first_line):
+            #     if not technical_comment:
+            #         technical_comment = comment
+            #     continue
```
smuzaffar, line 1216: @iarspider , better to delete this

process_pr.py
```diff
@@ -1476,11 +1534,11 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
     # end of parsing comments section
     cache_comment = None
-    if technical_comment:
-        cache_comment = technical_comment
+    if technical_comments:
+        cache_comment = technical_comments
     else:
         if already_seen:
-            cache_comment = already_seen
+            cache_comment = {0: already_seen}
```
smuzaffar, line 1541: @iarspider , once `technical_comments` is changed to `list` then please change this to `[already_seen]`

cms_static.py
```diff
@@ -11,7 +11,7 @@
 CMSBUILD_GH_USER = "cmsbuild"
 CMSBOT_IGNORE_MSG = "<cmsbot>\s*</cmsbot>"
 CMSBOT_NO_NOTIFY_MSG = "<notify>\s*</notify>"
-CMSBOT_TECHNICAL_MSG = "cms-bot internal usage"
+CMSBOT_TECHNICAL_MSG = "cms-bot internal usage(\d*)"
```
smuzaffar, line 14: @iarspider , I think we do not need this change. Technical comments will be in order any way. So no need to add `(index#)`

process_pr.py
```diff
@@ -230,31 +229,60 @@ def collect_commit_cache(bot_cache):
...
+        part, data = data[:65490], data[65490:]
+        res.append(part)
+
+    return res
+
+
+def write_bot_cache(bot_cache, cache_comments, issue, dryRun):
+    data = prepare_bot_cache(bot_cache)
+    for i, part in enumerate(data):
+        cache_comment = cache_comments[i]
```
smuzaffar, line 278: @iarspider , isn't `cache_comments[i]` going to fail for `len(data)>len(cache_comments)` ?

iarspider:
hold

iarspider:
Do not forget:
* Do not accidentally delete "already seen" comment!
* Can the order of technical comment and already seen comment even change? (I think not)

process_pr.py
```diff
@@ -230,31 +229,63 @@ def collect_commit_cache(bot_cache):
...
+
+
+def prepare_bot_cache(bot_cache):
+    res = []
+    data = dumps_maybe_compress(bot_cache)
+    while data:
+        # The limit is 65535 chars, and minimal bot cache comment is"cms-bot internal usage0<!-- bot cache:  -->",
+        # 43 characters. This leaves us 65492 chars for actual data. Keeping 2 extra characters in case bot cache
+        # spans more than 10 commits (usage0...usage9)
+        part, data = data[:65490], data[65490:]
```
smuzaffar, line 269: change this to 55K too

process_pr.py
```diff
@@ -1109,20 +1149,27 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
-            print("Read bot cache from already seen comment:", comment)
-        elif re.match(CMSBOT_TECHNICAL_MSG, first_line):
-            technical_comment = comment
-            bot_cache = read_bot_cache(comment_msg)
-            print("Read bot cache from technical comment:", comment)
+            bot_cache = extact_bot_cache(comment_msg)
+            if bot_cache:
+                print("Read bot cache from already seen comment:", comment)
+                technical_comments = [comment]
+                break
```
smuzaffar, line 1162: @iarspider , need to drop `break` here to make sure we process all comments in this loop

process_pr.py
```diff
@@ -1109,20 +1149,27 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
     # to use information during the actual comment loop
     for comment in issue.get_comments():
         all_comments.append(comment)
-        if bot_cache or technical_comment:
-            continue
         if comment.user.login.encode("ascii", "ignore").decode() != cmsbuild_user:
             continue
         comment_msg = comment.body.encode("ascii", "ignore").decode() if comment.body else ""
         first_line = "".join([l.strip() for l in comment_msg.split("\n") if l.strip()][0:1])
         if re.match(ISSUE_SEEN_MSG, first_line):
```
smuzaffar, line 1156: may be change this to 
```
if (not  already_seen) and re.match(ISSUE_SEEN_MSG, first_line):
```
to avoid running `re.match(ISSUE_SEEN_MSG, first_line)` once already seen message it found. There should not be more than one already seen messages and we are only interested in first one anyway

process_pr.py
```diff
@@ -1109,20 +1149,27 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
-            technical_comment = comment
-            bot_cache = read_bot_cache(comment_msg)
-            print("Read bot cache from technical comment:", comment)
+            bot_cache = extact_bot_cache(comment_msg)
+            if bot_cache:
+                print("Read bot cache from already seen comment:", comment)
+                technical_comments = [comment]
+                break
+        else:
+            if re.match(CMSBOT_TECHNICAL_MSG, first_line):
```
smuzaffar, line 1164: this can be `elif e.match(CMSBOT_TECHNICAL_MSG, first_line):` ... right?

iarspider:
unhold

process_pr.py
```diff
@@ -230,31 +229,63 @@ def collect_commit_cache(bot_cache):
...
+    for i, part in enumerate(data):
+        try:
+            cache_comment = cache_comments[i]
+        except (TypeError, IndexError):
+            cache_comment = None
+
+        old_body = cache_comment.body if cache_comment else CMSBOT_TECHNICAL_MSG
+        new_body = REGEX_COMMITS_CACHE.sub("", old_body) + "<!-- bot cache: " + part + " -->"
+        if old_body == new_body:
+            return
```
smuzaffar, line 286: @iarspider , this should be continue instead of return .... right?

process_pr.py
```diff
@@ -1109,20 +1149,25 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
         first_line = "".join([l.strip() for l in comment_msg.split("\n") if l.strip()][0:1])
-        if re.match(ISSUE_SEEN_MSG, first_line):
+        if (not already_seen) and re.match(ISSUE_SEEN_MSG, first_line):
             already_seen = comment
-            bot_cache = read_bot_cache(comment_msg)
-            print("Read bot cache from already seen comment:", comment)
+            bot_cache = extact_bot_cache(comment_msg)
+            if bot_cache:
+                technical_comments.append(comment)
+                continue
```
smuzaffar, line 1161: why not simplify it to 
```
if extact_bot_cache(comment_msg):
    technical_comments.append(comment)
```

and drop the `if not bot_cache:` condition at line 1165

smuzaffar: and may be then `extact_bot_cache` is not needed

process_pr.py
```diff
@@ -1109,20 +1149,25 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
-            bot_cache = read_bot_cache(comment_msg)
-            print("Read bot cache from technical comment:", comment)
+            technical_comments.append(comment)
+
+    if not bot_cache:
+        bot_cache = extract_bot_cache_v2(technical_comments)
+        print(
+            "Read bot cache from technical comment(s):",
+            ",".join(str(c) for c in technical_comments),
+        )
```
smuzaffar, line 1170: this can go in `extract_bot_cache_v2` itself

process_pr.py
```diff
@@ -1476,11 +1518,11 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
 
     # end of parsing comments section
     cache_comment = None
-    if technical_comment:
-        cache_comment = technical_comment
+    if technical_comments:
+        cache_comment = technical_comments
     else:
```
smuzaffar, line 1523: i think this part is not needed any more. `technical_comments` will have `already_seen` if `already_seen` contains the bot cache

process_pr.py
iarspider: I think it doesn't matter, unless we get extremely (un)lucky. The only case where it matters is if the cache is big enough to exceed one comment and one of the chunks of base64-encoded compressed cache is exactly the same as before, but overall base64-encoded compressed cache is not.

process_pr.py
smuzaffar: `extact_bot_cache` can be removed and we just directly use the following here  ... right?
```
    seen_commits_match = REGEX_COMMITS_CACHE.search(comment_msg)
    if seen_commits_match:
        technical_comments.append(comment)
```

process_pr.py
smuzaffar: if it does not matter then I would prefer to process all parts to make sure the contents are up to date

iarspider: Changed just in case.

process_pr.py
```diff
@@ -1109,20 +1147,19 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
-            print("Read bot cache from already seen comment:", comment)
+            if REGEX_COMMITS_CACHE.search(comment_msg):
+                technical_comments.append(comment)
         elif re.match(CMSBOT_TECHNICAL_MSG, first_line):
-            technical_comment = comment
-            bot_cache = read_bot_cache(comment_msg)
-            print("Read bot cache from technical comment:", comment)
+            technical_comments.append(comment)
+
+    if not bot_cache:
```
smuzaffar, line 1161: @iarspider , this condition is always true , so no need to check for this

process_pr.py
```diff
@@ -230,31 +229,61 @@ def collect_commit_cache(bot_cache):
...
-    old_body = cache_comment.body if cache_comment else CMSBOT_TECHNICAL_MSG
-    new_body = (
-        REGEX_COMMITS_CACHE.sub("", old_body)
-        + "<!-- bot cache: "
-        + dumps_maybe_compress(bot_cache)
-        + " -->"
+def extract_bot_cache(comment_msgs):
+    print(
+        "Reading bot cache from technical comment(s):",
+        ",".join(str(c) for c in technical_comments),
```
smuzaffar, line 245: @iarspider , ` technical_comments` should be `comment_msgs` here ... right?

process_pr.py
```diff
@@ -230,31 +229,61 @@ def collect_commit_cache(bot_cache):
...
+
+
+def prepare_bot_cache(bot_cache):
+    res = []
+    data = dumps_maybe_compress(bot_cache)
+    while data:
+        # The limit is 65535 chars, and minimal bot cache comment is"cms-bot internal usage0<!-- bot cache:  -->",
+        # 43 characters. This leaves us 65492 chars for actual data. Keeping 2 extra characters in case bot cache
+        # spans more than 10 commits (usage0...usage9)
+        part, data = data[:55000], data[55000:]
```
smuzaffar, line 267: @iarspider , update the comment above

smuzaffar:
ping

smuzaffar:
+externals

PR #2185, opened 2024-02-26
Add --maxmem_profile to all runTheMatrix workflows to log mamimum memory allocated

gartung:
addresses https://github.com/makortel/framework/issues/829
requires https://github.com/cms-sw/cmssw/pull/44001

### DIFF
--- a/run-ib-pr-matrix.sh
+++ b/run-ib-pr-matrix.sh
@@ -45,10 +45,11 @@ mkdir -p "$WORKSPACE/matrix-results"
 UC_TEST_FLAVOR=$(echo ${TEST_FLAVOR} | tr '[a-z]' '[A-Z]')
 pushd "$WORKSPACE/matrix-results"
   NJOBS=$(nproc)
-  CMD_OPTS="--maxmem_profile"
+  CMD_OPTS=""
   case "${TEST_FLAVOR}" in
     gpu )        MATRIX_ARGS="-w gpu ${MATRIX_ARGS}" ;;
     high_stats ) CMD_OPTS="-n 500" ; MATRIX_ARGS="-i all ${MATRIX_ARGS}" ;;
+    maxmem_profile ) CMD_OPTS="--maxmem_profile" ;;
     threading )  MATRIX_ARGS="-i all -t 4 ${MATRIX_ARGS}" ; let NJOBS=(${NJOBS}/4)+1 ;;
     nano )       MATRIX_ARGS="-w nano -i all ${MATRIX_ARGS}" ;;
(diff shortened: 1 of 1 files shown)


makortel:
> addresses [makortel/framework#673](https://github.com/makortel/framework/issues/673)

(this PR is more about https://github.com/makortel/framework/issues/829)

makortel:
I think there are some test flavors for which this monitoring would be harmful (like UBSAN, ASAN, `threading` etc). Perhaps the best way to get started would be to limit to the default testing set on x86 (we could then expand after gaining more experience).

antoniovilela:
https://github.com/cms-sw/cmssw/pull/44001 has been merged.

makortel:
Maybe we wait for Shahzad to get back to decide how exactly to continue (unless @aandvalenzuela @iarspider would know on the top of their head what exactly would need to be done to achieve https://github.com/cms-sw/cms-bot/pull/2185#issuecomment-1965167009).

run-ib-pr-matrix.sh
```diff
@@ -49,6 +49,7 @@ pushd "$WORKSPACE/matrix-results"
   case "${TEST_FLAVOR}" in
     gpu )        MATRIX_ARGS="-w gpu ${MATRIX_ARGS}" ;;
     high_stats ) CMD_OPTS="-n 500" ; MATRIX_ARGS="-i all ${MATRIX_ARGS}" ;;
+    maxmem_profile ) CMD_OPTS="--maxmem_profile" ;;
```
smuzaffar, line 52: @gartung , is `--maxmem_profile` supported by all cmssw release cycles? If not then I would suggest to run `runTheMatrix.py --help` to see if this opt is supported or not. I n case a cmssw release cycle does not support it then just mark the commit status successfull with message `Memory profile not run due to lack of support`

smuzaffar:
So remind me what we want to do with this change? from the change it looks like we want to run extra relvals with `--maxmem_profile` option e.g. one can say `enable maxmem_profile` and then bot should run normal relval + all normal relvals with `--maxmem_profile` .. right?

makortel:
I want all normal relval on x86 (only) to be run with `--maxmem_profile` (i.e. no additional "test flavor"). Any extra workflows could include the `--maxmem_profile` too, but for other test flavors (non-x86 architectures, different IB flavor, extra tests) for now.

smuzaffar:
@gartung @makortel I have opened https://github.com/cms-sw/cms-bot/pull/2202 to enable memory profiling for default workflows for production IB/arch only

makortel:
I think we can close this PR in favor of #2202

makortel:
@cmsbuild, please close

PR #2190, opened 2024-03-01
SCP dummy file to /afs cmsbuild area for aarch64 or ppc64le blacklisted nodes

aandvalenzuela:
Having a file in `/afs` with the status of the node sanity check can help to automatize cleanup via cron job, e.g., running cvmfs cleanup on olarm machines automatically


aandvalenzuela:
Nodes status can be found at `/afs/cern.ch/user/c/cmsbuild/nodes-info/<node-name>` as either PASS or ERROR. Cron job will check and run only if status is ERROR.
Sanity check runs as part of https://cmssdt.cern.ch/jenkins/job/nodes-sanity-check/

In the same way, files under `/afs/cern.ch/user/c/cmsbuild/disk-info/<node-name>` contain disk space left on the node (if space is less than 10GB). For the time being, it checks only for `olarm` nodes (but more nodes can be added). The idea would be to automatically clean the nodes that have less than 10GB of space left (or blacklist them).
Disk check runs as part of https://cmssdt.cern.ch/jenkins/job/jenkins-infrastructure-sanity-check/

PR #2196, opened 2024-03-14
Limit how many files can be changed in a PR

iarspider:


process_pr.py
```diff
@@ -750,13 +754,20 @@ def get_changed_files(repo, pr, use_gh_patch=False):
         print("PR Files: ", pr_files)
         return pr_files
     cmd = (
-        "curl -s -L https://patch-diff.githubusercontent.com/raw/%s/pull/%s.patch | grep '^diff --git ' | sed 's|.* a/||;s|  *b/| |' | tr ' ' '\n' | sort | uniq"
+        "curl -s -L https://patch-diff.githubusercontent.com/raw/%s/pull/%s.diff | grep '^diff --git ' | sed 's|.* a/||;s|  *b/| |' | tr ' ' '\n' | sort | uniq"
         % (repo.full_name, pr.number)
     )
     e, o = run_cmd(cmd)
     if e:
         return []
```
smuzaffar, line 762: @iarspider , we should also let bot fail on error `e` ... right?

PR #2214, opened 2024-04-18
process_pr: only trigger automatic tests for first L2 signature

iarspider:


smuzaffar:
@missirol , this should fix the issue of re-triggering PR tests ( as we have seen for https://github.com/cms-sw/cmssw/pull/44735 ) with one side effect see https://github.com/cms-sw/cms-bot/pull/2214#issuecomment-2063585513

missirol:
Thank you @smuzaffar !

PR #2221, opened 2024-05-02
ignore ref wf failures if Pr tests did not run that wf

smuzaffar:
This should fix the issue reported [here](https://github.com/cms-sw/cmssw/pull/44767#issuecomment-2087482148). There was a typo (`-gt` instead of `-eq`) and bot should have ignored the extra failed workflows ran in the baseline.


smuzaffar:
FYI @makortel

makortel:
Thanks!

PR #2230, opened 2024-05-14
Add script to run HLT timing test

iarspider:


smuzaffar:
@iarspider , I would suggest to make use of `run-ib-testbase.sh` for creating and setting up cmssw dev area e.g. see https://cmssdt.cern.ch/jenkins/job/ib-run-addons/configure

PR #2246, opened 2024-05-29
Add `PhysicsTools/XGBoost` to ml category

mmusich:
Needed by https://github.com/cms-sw/cmssw/pull/45085


mmusich:
@cms-sw/ml-l2 FYI

PR #2247, opened 2024-05-29
Use different paths for each LUMI slot in Jenkins

aandvalenzuela:
Using a different `/scratch/$SLURM_ACCOUNT/$USER/$SESSION` path for each OS. Create CVMFS cache and scratch area if needed.


aandvalenzuela:
Merging since it only changes the paths to make a user/session distinction on the node

PR #2256, opened 2024-06-06
Process pr unit tests 3

iarspider:


iarspider:
@smuzaffar this is with PyGithub version frozen to 1.56 (last one that supports Python 3.6).

process_pr.py
```diff
@@ -463,10 +476,7 @@ def has_user_emoji(bot_cache, comment, repository, emoji, user):
         e = bot_cache["emoji"][comment_id]
     else:
         emojis = None
-        if "Issue.Issue" in str(type(comment)):
-            emojis = get_issue_emojis(comment.number, repository)
-        else:
-            emojis = get_comment_emojis(comment.id, repository)
+        emojis = comment.get_reactions()
```
smuzaffar, line 479: Does this work for PR description comment?

process_pr.py
```diff
@@ -1522,32 +1550,30 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
-                    issue.create_comment(
-                        "This PR contains many commits ({0} >= {1}) and will not be processed. "
-                        "Please ensure you have selected the correct target branch and consider squashing unnecessary commits.\n"
-                        "{2}, to re-enable processing of this PR, you can write `+commit-count` in a comment. Thanks.".format(
-                            pr.commits,
-                            TOO_MANY_COMMITS_WARN_THRESHOLD,
-                            ", ".join([gh_user_char + name for name in CMSSW_ISSUES_TRACKERS]),
-                        )
-                    )
+                issue.create_comment(
```
smuzaffar, line 1553: why remove the `if not dryRun:` check here? For normal run this will create the comment ... right?

process_pr.py
```diff
@@ -1522,32 +1550,30 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
...
-                if not dryRun:
-                    issue.create_comment(
-                        "This PR contains too many commits ({0} >= {1}) and will not be processed.\n"
-                        "Please ensure you have selected the correct target branch and consider squashing unnecessary commits.\n"
-                        "The processing of this PR will resume once the commit count drops below the limit.".format(
-                            pr.commits,
-                            TOO_MANY_COMMITS_FAIL_THRESHOLD,
-                        )
-                    )
+                issue.create_comment(
```
smuzaffar, line 1563: same here? should we keep the dryRun check on?

process_pr.py
iarspider: This is why I wanted to always overload Pygithub methods...

PR #2257, opened 2024-06-07
Override local SITECONF if needed

aandvalenzuela:
Useful for LUMI nodes where SITECONF should come from `T2_FI_HIP`


smuzaffar:
please test

PR #2259, opened 2024-06-13
Fix duplicate key in categories.py

iarspider:


smuzaffar:
+externals
thanks @iarspider . Is there any way we can check this via ci tests?

PR #2261, opened 2024-06-14
Improve generate-categories-json.py

iarspider:
@smuzaffar let me know if I should've only changed hashbang and removed import


smuzaffar:
looks good

PR #2275, opened 2024-06-30
Remove downloaded RPMs before publishing

iarspider:
Some RPMs (e.g. cms+data-SimG4CMS-Calo+V03-05-00-1-1.el9_amd64_gcc12.rpm) are larger than 2M limit, causing CVMFS publication to fail.

### DIFF
--- a/cvmfsInstall.sh
+++ b/cvmfsInstall.sh
@@ -147,7 +147,6 @@ for REPOSITORY in $REPOSITORIES; do
         ${CMSPKG} install -y $x || true
         time ${CMSPKG} install -y `echo $x | sed -e 's/cmssw-ib/cmssw/'` || true
         time ${CMSPKG} install -y `echo $x | sed -e 's/cmssw-ib/cmssw-patch/'` || true
-        rm -rf $WORKDIR/*/var/cmspkg/rpms || true
         relname=`echo $x | awk -F + '{print $NF}'`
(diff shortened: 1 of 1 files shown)


cvmfsInstall.sh
```diff
@@ -188,6 +188,8 @@ rm -f $BASEDIR/latest
 ln -s $(grep "^nweek-" ${CMS_BOT_DIR}/ib-weeks | tail -1) $BASEDIR/latest
 
 if $CVMFS_INSTALL ; then
+  # Remove RPMs before publishing
+  rm -rf $WORKDIR/*/var/cmspkg/rpms || true
```
smuzaffar, line 192: @iarspider , any idea why https://github.com/cms-sw/cms-bot/blob/master/cvmfsInstall.sh#L150 is not enough?

iarspider: For some jobs (the ones triggered by cmsrep webhook) we do not pass `RELEASE_FORMAT` (which is assigned to `RELEASE_NAME`), thus the condition at [line 144](https://github.com/cms-sw/cms-bot/blob/master/cvmfsInstall.sh#L144) evaluates to false. I guess I can move cleanup at line 150 below [line 160](https://github.com/cms-sw/cms-bot/blob/master/cvmfsInstall.sh#L160) so that it is always executed...

PR #2278, opened 2024-07-01
Add L1TriggerScouting/OnlineProcessing to DAQ and cleanup old packages

smorovic:
Adding new package to `daq` supervision.

Package is submitted to CMSSW here:
https://github.com/cms-sw/cmssw/pull/45350

Old DAQ packages that no longer exist in the CMSSW codebase (Run1 period) are cleaned up.

### DIFF
--- a/categories_map.py
+++ b/categories_map.py
@@ -374,6 +374,7 @@ CMSSW_CATEGORIES = {
         "EventFilter/StorageManager",
         "EventFilter/Utilities",
         "L1TriggerScouting/Utilities",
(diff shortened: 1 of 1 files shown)


smuzaffar:
@smorovic , we can not remove/cleanup old packages as old releases still contain these and PRs for old releases need these

smorovic:
OK, I reverted that commit.

There is little probability anything there will need to be updated (old DAQ event filter code), but there is no real need to remove references either.

smuzaffar:
+externals

PR #2279, opened 2024-07-01
Relax chk-invalid-headers checks to match SCRAM rules:

iarspider:
- file in PackageA/SubpackageA can include any file from that subpackage, but only interface (public header) from other package or subpackage
- file in PackageA/SubpackageA/test can, in addition, include file from any other package's `test` directory

### DIFF
--- a/chk-invalid-headers.py
+++ b/chk-invalid-headers.py
@@ -56,7 +56,9 @@ def main():
             continue
         for src in usedby[inc].split(" "):
             sitems = src.split("/")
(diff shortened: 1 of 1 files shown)


smuzaffar:
looks good

makortel:
> * file in PackageA/SubpackageA can include any file from that subpackage, but only interface (public header) from other package or subpackage
> * file in PackageA/SubpackageA/test can, in addition, include file from any other package's `test` directory

These seem to be more relaxed than our policy (6.4 in https://cms-sw.github.io/cms_coding_rules.html#6--packaging-rules-1). Am I missing something?

PR #2280, opened 2024-07-02
Remove data workflows from profiling in PRs

jfernan2:
It is known[1] that igprof time profiling crashes at step3 (RECO) for data workflows due to an unknown issue with TF in DeepTau ID[2].

To avoid the problem and at least have profilings for MC workflows which are the most interesting in view of the Phase2 SW upgrades, this PR removes data workflows from the test. Besides, profiling scripts used for PR profiling have been recently updated to only monitor up to step3 (RECO) [3] (thanks @gartung !!)

[1] https://github.com/cms-sw/cmssw/issues/43166
[2] https://github.com/cms-sw/cmssw/issues/42444
https://github.com/cms-sw/cmssw/issues/28358
[3] https://github.com/cms-cmpwg/profiling.git


smuzaffar:
+externals

PR #2291, opened 2024-07-10
add Mario to tracking POG

slava77:
@mmasciov


smuzaffar:
+externals

@mmasciov added as `tracking` POG by @cms-sw/tracking-pog-l2

PR #2292, opened 2024-07-11
Allow closing/reopening a PR even if there are too many commits or changed files

iarspider:


smuzaffar:
looks good @iarspider. Although it should work but do we have any old cmssw PR with a lot of commits or files which we can use to test this change?

smuzaffar:
+externals

though it did not work for https://github.com/cms-sw/cmssw/pull/45400 as the original branch has been deleted but at least bot properly tried to open the PR

PR #2295, opened 2024-07-16
Update gpu workflows for CMSSW 14.x

fwyzard:
Update the GPU workflows for CMSSW 14.0 and later to run the Alpaka-based configurations:
  - Pixel only, with GPU-vs-CPU validation
  - ECAL only
  - HCAL only, with GPU-vs-CPU validation


fwyzard:
enable gpu

fwyzard:
please test

smuzaffar:
+externals

PR #2301, opened 2024-07-24
clean up some heavy-ion users that long ago left CMS

mandrenguyen:
These folks no longer work in CMS since many years.
Although it's not hurting anyone that they get pinged, it's a bit annoying when you open a PR, as it's harder to see whether the people who are actually needed are alerted.


mandrenguyen:
+1

mandrenguyen:
+externals
Just checking whether I have these rights.  Presumably not, or it the plus one would have triggered it.

PR #2302, opened 2024-07-30
Fix installation of RISC-V IBs

iarspider:
ib-install-cvmfs fails for riscv IBs: `cvmfsInstall.sh: line 63: /cvmfs/cms-ib.cern.ch/sw/riscv64/stratum0: No such file or directory`


smuzaffar:
-externals

We can not deploy riscv64 rpms yet. We need to have build a fc38/riscv64 container first and then use it via QEMU

PR #2305, opened 2024-08-03
Update categories_map.py to assign category for GeometryAligner

bsunanda:
The GeometryAligner.h has to be moved from Geometry/CommonTopologies to avoid unusual dependencies


bsunanda:
@cmsbuild Please test

PR #2307, opened 2024-08-09
Update gh-teams.py, changing GEN team

bbilin:


mandrenguyen:
+1

smuzaffar:
@bbilin , no need to explicitly update `generator-l2` admins here. All the members of @cms-sw/generators-l2 (from https://github.com/cms-sw/cms-bot/blob/master/categories.py#L66-L83 ) are already there

PR #2309, opened 2024-08-13
Change default profiling workflow searched for

gartung:
@makortel @smuzaffar this is why the Resources Piecharts links are not being added to the IB dashboard.


gartung:
please test

gartung:
please test

PR #2323, opened 2024-08-27
[RFC] Disable checking for number of modified files in non-cmssw repos

iarspider:
@smuzaffar I think this change makes sense - we only need to check what files were touched to determine what L2 signatures the PR needs, right?


iarspider:
hold

PR #2338, opened 2024-09-20
Region parameter was not respected - also usage typos

oljemark:
Hello. The --region parameter was updating the wrong, misspelled variable. Also some typos in the usage printout.


smuzaffar:
+externals
thanks

PR #2342, opened 2024-09-24
avoid duplication in l2 list

kpedro88:
I noticed in https://github.com/cms-sw/cmssw/issues/46113 that the list of tagged L2s could contain duplicates when multiple categories are assigned.

I chose the most broadly compatible approach (works across Python 2 and 3) to remove the duplicates, preserving the order to be as consistent as possible with previous behavior. Tested in standalone Python and it works as expected.

### DIFF
--- a/process_pr.py
+++ b/process_pr.py
@@ -2189,6 +2189,7 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
             for signature in new_assign_cats
             if signature in l2_categories
         ]
+        new_l2s = sorted(set(new_l2s), key=lambda s: s.lower())
         if not dryRun:
             issue.create_comment(
                 "New categories assigned: "
--- a/tests/PRActionData/TestProcessPr.test_assign.json
+++ b/tests/PRActionData/TestProcessPr.test_assign.json
@@ -80,7 +80,7 @@
     },
     {
         "type": "create-comment",
-        "data": "New categories assigned: dqm,db\n\nrvenditti,syuvivida,tjavaid,nothingface0,antoniovagnerini,francescobrivio,saumyaphor4252,perrotta,consuegs you have been requested to review this Pull request/Issue and eventually sign? Thanks"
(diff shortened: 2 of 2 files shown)


process_pr.py
```diff
@@ -2189,6 +2189,7 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
             for signature in new_assign_cats
             if signature in l2_categories
         ]
+        new_l2s = list(OrderedDict.fromkeys(new_l2s))
```
smuzaffar, line 2192: thanks @kpedro88 . I guess just using `sorted(set(new_l2s))` should work too ... right ? I guess we do not care about the order too much here so having a sorted list will be good too

kpedro88: Sure, I can change to that if the order doesn't matter.

smuzaffar: +1
yes order here does not matter.

kpedro88:
@smuzaffar I updated a test input file manually because I had a failed workflow run from https://github.com/cms-sw/cms-bot/blob/master/.github/workflows/test-process-pr.yaml. However, updating the input file doesn't trigger the workflow and it can't be dispatched manually, so I can't confirm it's fixed. I also don't know if `tests/ReplayData` should be similarly updated.

tests/PRActionData/TestProcessPr.test_assign.json
```diff
@@ -80,7 +80,7 @@
     },
     {
         "type": "create-comment",
-        "data": "New categories assigned: dqm,db\n\nrvenditti,syuvivida,tjavaid,nothingface0,antoniovagnerini,francescobrivio,saumyaphor4252,perrotta,consuegs you have been requested to review this Pull request/Issue and eventually sign? Thanks"
+        "data": "New categories assigned: dqm,db\n\nantoniovagnerini,consuegs,francescobrivio,nothingface0,perrotta,rvenditti,saumyaphor4252,syuvivida,tjavaid you have been requested to review this Pull request/Issue and eventually sign? Thanks"
```
smuzaffar, line 83: @iarspider , any suggestions? should we update the test file directly here or do you want to auto generate it?

iarspider: @smuzaffar yes, we can update it manually.

smuzaffar:
+externals

PR #2346, opened 2024-10-01
Add script to get file category

iarspider:


smuzaffar:
why not use or update https://github.com/cms-sw/cms-bot/blob/master/package2category.py  ?

PR #2360, opened 2024-10-16
Update of L2 PRO POG conveners

forthommel:
Removing Ksenia/Michael from PRO POG conveners, and adding Andrea/Laurent instead.


smuzaffar:
@michael-pitt , @kshcheli , @AndreaBellora , I hope you are fine with this change?

PR #2361, opened 2024-10-18
PR Testing: check for warnings also in files not directly touched by PR

iarspider:
As suggested here: https://github.com/cms-sw/cmssw/pull/46388#issuecomment-2421667896

### DIFF
--- a/pr_testing/test_multiple_prs.sh
+++ b/pr_testing/test_multiple_prs.sh
@@ -19,7 +19,7 @@ function get_compilation_warnings() {
 
 function get_warnings_files(){
   for i in $(cat $1 | sed 's|^src/||;s|:.*||;s| ||g;s|[(].*||' | sort -u) ; do
-    [ $(grep "$i" $WORKSPACE/changed-files | wc -l) -gt 0 ] && echo $i
+    [ $(grep "$i" $WORKSPACE/full-list-of-changed-files.txt | wc -l) -gt 0 ] && echo $i
   done
 }
 
@@ -70,9 +70,7 @@ function process_changed_files() {
     # Call the function to extract filenames and append them atomically
     extract_filenames "$headername"
   done < "$WORKSPACE/changed-files"
-  local temp_file=$(mktemp)
-  sort -u $WORKSPACE/changed-files $WORKSPACE/indirectly-changed-files > $temp_file
-  mv $temp_file $WORKSPACE/changed-files
+  sort -u $WORKSPACE/changed-files $WORKSPACE/indirectly-changed-files > $WORKSPACE/full-list-of-changed-files.txt
 }
 
 # Constants
--- a/pr_testing/test_multiple_prs.sh
+++ b/pr_testing/test_multiple_prs.sh
@@ -44,6 +44,37 @@ function get_pr_relval_args() {
   echo "${WF_ARGS}"
 }
 
+# Function to extract filenames by headername and append to indirectly-changed-files
+function extract_filenames() {
+  local headername="$1"
+  local input_file="./etc/dependencies/usedby.out"
+  local output_file="$WORKSPACE/indirectly-changed-files"
+  [ ! -e $output_file ] && touch $output_file
+
+  # Extract lines starting with headername, split them, and append each filename to the temp file
+  grep "^$headername" "$input_file" | while read -r line; do
+    # Split the line into an array
+    IFS=' ' read -r -a array <<< "$line"
+
+    # Loop through each element after the first (which is the headername)
+    for filename in "${array[@]:1}"; do
+      echo "$filename" >> "$output_file"
+    done
+  done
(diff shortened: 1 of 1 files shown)


iarspider:
@smuzaffar this PR is ready.

smuzaffar:
test parameters:
- addpkg = FWCore

smuzaffar:
please test

iarspider:
please test

pr_testing/test_multiple_prs.sh
```diff
@@ -44,6 +44,37 @@ function get_pr_relval_args() {
...
+function process_changed_files() {
+  [ -e "$WORKSPACE/indirectly-changed-files" ] || touch "$WORKSPACE/indirectly-changed-files"
+  # Iterate over each line in $WORKSPACE/changed-files
+  while IFS= read -r headername; do
+    # Call the function to extract filenames and append them atomically
+    extract_filenames "$headername"
+  done < "$WORKSPACE/changed-files"
+  local temp_file=$(mktemp)
+  sort -u $WORKSPACE/changed-files $WORKSPACE/indirectly-changed-files > $temp_file
+  mv $temp_file $WORKSPACE/changed-files
```
smuzaffar, line 75: @iarspider , we are overriding the original `$WORKSPACE/changed-files` file. I thought we discussed that we should create a separate file for direct + indirect change files

smuzaffar: we discussed that as `changed-files` is also used for other purposes so it is safer to use a different file for warnings. So please create a separate file. Better to name it such that it end up with `.txt` extension so that https://github.com/cms-sw/cms-bot/blob/master/pr_testing/_helper_functions.sh#L82 can upload it ( just for debugging purposes)

smuzaffar: > You suggested to pre-fill changed-files with directly changed files, then update it after running gcc compilation
 
correct, it should have both direct and indirect file but it should be in different file than `changed-files`

iarspider:
please test

pr_testing/test_multiple_prs.sh
```diff
@@ -1141,6 +1119,16 @@ GENERAL_ERRORS=`grep "ALL_OK" $WORKSPACE/build.log` || true
 
 rm -f $WORKSPACE/deprecated-warnings.log
 get_compilation_warnings $WORKSPACE/build.log > $WORKSPACE/all-warnings.log
+
+pushd ..
+scram b echo_SCRAM_TOOL_HOME
+test -e config/SCRAM/findDependencies.py
+SCRAM_TOOL_HOME=`scram b echo_SCRAM_TOOL_HOME 2>/dev/null | tail -1 | cut -d' ' -f3`
+mkdir -p etc/dependencies
+SCRAM_TOOL_HOME=$SCRAM_TOOL_HOME ./config/SCRAM/findDependencies.py -rel `pwd` -arch ${SCRAM_ARCH}
```
smuzaffar, line 1128: @iarspider , this will not work for old perl based releases . See https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_11_0_X/master/scram-project-build.file#L210 . So you need to check if `config/scram_version` has scram version V3_* then use findDependencies.py otherwise use findDependencies.pl ( with correct parameters)

pr_testing/test_multiple_prs.sh
```diff
@@ -44,6 +44,35 @@ function get_pr_relval_args() {
...
+    # Loop through each element after the first (which is the headername)
+    for filename in "${array[@]:1}"; do
+      echo "$filename" >> "$output_file"
+    done
+  done
+}
+
+# Function to get indirectly changed files
+function process_changed_files() {
+  [ -e "$WORKSPACE/indirectly-changed-files" ] || touch "$WORKSPACE/indirectly-changed-files"
```
smuzaffar, line 67: just `touch "$WORKSPACE/indirectly-changed-files"` should be enough

pr_testing/test_multiple_prs.sh
iarspider: Noted.

smuzaffar: @iarspider , we still need to fix this perl based scram

iarspider:
please test

iarspider:
please test

smuzaffar:
please test for CMSSW_10_6_X

iarspider:
please test for CMSSW_10_6_X

iarspider:
please test for CMSSW_10_6_X

smuzaffar:
+externals

looks good

smuzaffar:
@iarspider , looks like this broke the Pr tests. The static checks are running `scram b clean` which basically deletes allthe generated products. We need to either revert this or need a quick fix

PR #2377, opened 2024-11-18
Set TMPDIR in docker_launcher.sh if not set already

iarspider:
This should fix Risc-V builds. @smuzaffar as discussed.


iarspider:
please test

smuzaffar:
externals

`docker_launcher.sh` did not fail, so changes should be fine for non-riscv64. In order to test it properly, we need to get the riscv64 IB.

PR #2384, opened 2024-11-25
Parse package info data and push it to OpenSearch

aandvalenzuela:
Parse package info data and push it to OpenSearch.


ib-pkg-info.py
```diff
@@ -0,0 +1,82 @@
...
+        for package in packages:
+            payload = {
+                "ib_name": package_file.split("/")[6],
+                "release_cycle": release_cycle,
+                "flavor": flavor,
+                "date": date,
+                "architecture": architecture,
+                "@timestamp": get_current_time(),
+                package: packages[package],
+                "jenkins_server": JENKINS_PREFIX,
```
smuzaffar, line 71: @aandvalenzuela , what is the purpose of sending `jenkins_server` ? I think we do not need it

smuzaffar:
why we need two scripts ?

aandvalenzuela:
Hi @smuzaffar, I have merged the scripts into one but I haven't found IB data on cmsrep.

PR #2389, opened 2024-12-03
Disable auto-tests

iarspider:


iarspider:
@smuzaffar

smuzaffar:
looks good, lets merge it but we should re-enable it for PR with squash commits (where change before and after the squash remains same and at least one l2 has already signed it)

PR #2396, opened 2024-12-08
Add "ngt" label

fwyzard:
Add the `ngt` label, to support the `type ngt` bot command.

### DIFF
--- a/githublabels.py
+++ b/githublabels.py
@@ -41,6 +41,7 @@ TYPE_COMMANDS = {
         "performance|improvements|performance-improvements",
         "mtype",
     ],
+    "ngt": [LABEL_COLORS["performance"], "ngt", "mtype"],
 }
 
 for lab in get_dpg_pog():


fwyzard:
please test

fwyzard:
type ngt

(I'm just curious to see what happens)

smuzaffar:
@fwyzard , what is `ngt` ? 
If it is only used for adding `ngt` label by explicitly using `type ngt` comment then please include it in https://github.com/cms-sw/cms-bot/blob/master/githublabels.py#L32 . If we want bot to automatically set this label based on changes set then it should go in https://github.com/cms-sw/cms-bot/blob/master/categories_map.py#L1801

fwyzard:
> what is `ngt` ?

the NextGen Trigger project, that is developing additional L1T and HLT functionalities for Phase 2.

The idea is to be able to quickly point out which developments come from this project, since we have to give yearly reports on the developments to the foundations that funds the project. 

> If it is only used for adding `ngt` label by explicitly using `type ngt` comment then please include it in https://github.com/cms-sw/cms-bot/blob/master/githublabels.py#L32

Yes, that was the idea.
I'll make the changes, thanks for the correction.

githublabels.py
```diff
@@ -41,6 +41,7 @@ def get_dpg_pog():
         "performance|improvements|performance-improvements",
         "mtype",
     ],
+    "ngt": [LABEL_COLORS["performance"], "ngt", "mtype"],
```
fwyzard, line 44: @smuzaffar is this what you had in mind ?

smuzaffar: yes.

fwyzard:
please test

smuzaffar:
+externals

PR #2413, opened 2025-01-14
Add HeterogeneousCore/AlpakaMath package to heterogeneous

makortel:
Package is added in https://github.com/cms-sw/cmssw/pull/47033


makortel:
@cms-sw/externals-l2

smuzaffar:
+externals

Issue #2417, opened 2025-01-20
keep record stack trace of failed relvals/unit tests in opensearch


smuzaffar:
For each relval, we keep trace of all `%MSG-e ...` messages and cms exceptions. We should also keep trace of stack trace (if available). For example a failed relvals with output like
```
#0  0x000040003def6960 in poll () from /lib64/libc.so.6
#1  0x0000400042fb5100 in edm::service::InitRootHandlers::stacktraceFromThread() () from /cvmfs/cms-ib.cern.ch/sw/aarch64/nweek-02872/el9_aarch64_gcc12/cms/cmssw/CMSSW_15_0_X_2025-01-13-2300/lib/el9_aarch64_gcc12/pluginFWCoreServicesPlugins.so
#2  0x0000400042fb5334 in sig_dostack_then_abort () from /cvmfs/cms-ib.cern.ch/sw/aarch64/nweek-02872/el9_aarch64_gcc12/cms/cmssw/CMSSW_15_0_X_2025-01-13-2300/lib/el9_aarch64_gcc12/pluginFWCoreServicesPlugins.so
#3  <signal handler called>
#4  0x0000400131ffe038 in ?? ()
#5  0x00004000514463b0 in cling::Interpreter::RunFunction(clang::FunctionDecl const*, cling::Value*) () from /cvmfs/cms-ib.cern.ch/sw/aarch64/nweek-02872/el9_aarch64_gcc12/cms/cmssw/CMSSW_15_0_X_2025-01-13-2300/external/el9_aarch64_gcc12/lib/libCling.so
#6  0x00004000514463b0 in cling::Interpreter::RunFunction(clang::FunctionDecl const*, cling::Value*) () from /cvmfs/cms-ib.cern.ch/sw/aarch64/nweek-02872/el9_aarch64_gcc12/cms/cmssw/CMSSW_15_0_X_2025-01-13-2300/external/el9_aarch64_gcc12/lib/libCling.so
#7  0x000040005144751c in cling::Interpreter::EvaluateInternal(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, cling::CompilationOptions, cling::Value*, cling::Transaction**, unsigned long) () from /cvmfs/cms-ib.cern.ch/sw/aarch64/nweek-02872/el9_aarch64_gcc12/cms/cmssw/CMSSW_15_0_X_2025-01-13-2300/external/el9_aarch64_gcc12/lib/libCling.so
[...]

PR #2419, opened 2025-01-23
Send stacktrace to es

iarspider:
Solves https://github.com/cms-sw/cms-bot/issues/2417

### DIFF
--- a/es_relval_log.py
+++ b/es_relval_log.py
@@ -128,6 +128,8 @@ def es_parse_log(logFile):
     errors = []
     inException = False
     inError = False
+    inStacktrace = False
+    stacktrace = []
     datasets = []
     error_count = 0
     if exists(logFile):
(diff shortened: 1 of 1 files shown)


smuzaffar:
thanks @iarspider , looks good to me

iarspider:
As an alternative, we can store stacktrace as a list of function names, without frame numbers (`#3`, `#4`, etc) and without library names (`from /cvmfs/cms-ib.cern.ch/sw/aarch64/...`), it can make searching easier. We can also store the name of first named function in a separate key, since it's usually the most interesting part.

smuzaffar:
I think current implemetation is more than enough. One can always search `stacktrace:*some-string*` to find all relvals containing it

smuzaffar:
it will be nice to add this for unit tests too https://github.com/cms-sw/cms-bot/blob/master/es_ibs_log.py#L152-L207

PR #2431, opened 2025-02-04
cmsBuild from V00-33 and above support --define option

smuzaffar:
This should fix the PR tests for old release cycles which use pkgtools V00-31-XX where `cmsBuild` does not support `--define` command line option


smuzaffar:
looks good for cms-sw/cmsdist#9673 tests

PR #2451, opened 2025-03-14
Use python logging instead of print in process_pr.py

iarspider:
Also update python version used for running tests locally (github workflow doesn't use that script).


iarspider:
hold
Need more unit tests

PR #2461, opened 2025-03-20
Extra tests: accept 'high_stats', not 'high-stats'

iarspider:
<cmsbot></cmsbot>
We can't just replace dash with underscode, since we use dash in `multi-microarchs`. If you think that we should use underscoes everywhere, I can do it.


iarspider:
enable high-stats

iarspider:
enable high_stats

smuzaffar:
+externals

I think we can merge it so that bot can recognize it for other PR

PR #2465, opened 2025-03-25
process_pr: add tests for test options

iarspider:
Tests both `please test` and `test parameters` comments. Changes to `process_pr` are needed to sort enabled tests and relvals - alternatively I can sort these values when loading and comparing actions.


smuzaffar:
test parameters:
- skip_test = static, header
- enable = gpu threading, nano
- addpkg = HeterogeneousTest , FWCore

process_pr.py
```diff
@@ -630,6 +630,7 @@ def check_matrix_extras(first_line, repo, params, mkey, param, *args):
     print(first_line, repo, params, mkey, param)
     if kitem[-1] in EXTRA_RELVALS_TESTS:
         param = param + "_" + kitem[-1].upper().replace("-", "_")
+    first_line = ",".join(sorted(first_line.split(",")))
```
smuzaffar, line 633: @iarspider , should we drop trim (removing space char) each item too?

process_pr.py
```diff
@@ -1560,7 +1561,7 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
                     test_comment = comment
                     abort_test = None
                     cmssw_prs = v2
-                    extra_wfs = v3
+                    extra_wfs = ",".join(sorted(v3.split(",")))
```
smuzaffar, line 1564: same here, should we trim each item?

process_pr.py
iarspider: We can, or we can wait until next PR, in which we will normalize whitespaces, remove `@cmsbuild please` prefix and remove spaces around commas.

smuzaffar:
+externals

PR #2468, opened 2025-04-07
IB relvals: get number of supported GPUs instead of hardcoding it

iarspider:


iarspider:
@fwyzard FYI

jobs/jobscheduler.py
```diff
@@ -70,8 +71,23 @@ def runJob(job):
...
+        if job.get("gpu", -1) >= 0:
+            maker, idx = job["gpu"].split(":", 1)
+            # Remove assignment of GPUs if it exists
+            cmd = re.sub(r"HIP_VISIBLE_DEVICES=\S+", "", cmd)
+            cmd = re.sub(r"CUDA_VISIBLE_DEVICES=\S+", "", cmd)
+
+            if maker == "cuda":
+                cmd = cmd.replace(
+                    "cmsDriver.py",
+                    "CUDA_VISIBLE_DEVICES=%s HIP_VISIBLE_DEVICES=0 cmsDriver.py" % idx,
```
fwyzard, line 83: I'm not sure what this does, but keep in mind that `0` means "use the first GPU", not that there are 0 GPUs.
For that, pass an empty value.

jobs/jobscheduler.py
```diff
@@ -215,8 +231,7 @@ def initJobs(jobs, resources, otype):
             total_jobs += 1
             job = group["commands"][i]
             job["origtime"] = job["time"]
-            if "HIP_VISIBLE_DEVICES=0" in job["command"]:
-                job["gpu"] = -1
+            job["gpu"] = 1 if "-w gpu" in job["command"] else -1
```
smuzaffar, line 234: @iarspider , `-w gpu` is runTheMatrix option. The `command` here is `cmsDriver.py` and it will not have this option

jobs/jobscheduler.py
```diff
@@ -70,8 +71,23 @@ def runJob(job):
...
+            if maker == "cuda":
+                cmd = cmd.replace(
+                    "cmsDriver.py",
+                    "CUDA_VISIBLE_DEVICES=%s HIP_VISIBLE_DEVICES= cmsDriver.py" % idx,
+                )
+            elif maker == "rocm":
+                cmd = cmd.replace(
+                    "cmsDriver.py",
+                    "HIP_VISIBLE_DEVICES=%s CUDA_VISIBLE_DEVICES= cmsDriver.py" % idx,
+                )
```
smuzaffar, line 89: @fwyzard , should we support `CUDA_VISIBLE_DEVICES=N HIP_VISIBLE_DEVICES=N` or will we only request one type of gpu? I mean in future can runTheMatrix request to run a workflow step to use both amd and nvidia gpus?

jobs/jobscheduler.py
smuzaffar: you can see https://cmssdt.cern.ch/SDT/cgi-bin/logreader/el8_amd64_gcc12/CMSSW_15_1_ROCM_X_2025-04-06-2300/pyRelValMatrixLogs/run/141.044406_Run3-2023_JetMET2023D_RecoPixelOnlyTripletsGPU/job.json#/ for the details of `job` object

iarspider: I see. I'll find a different solution.

cmsutils.py
```diff
@@ -70,6 +70,31 @@ def _memorySizeGB():
...
+    else:
+        raise ValueError("Unsupported GPU type: {0}".format(gpu_type))
+
+    e, o = run_cmd(cmd)
+
+    if e != 0:
+        return 0
+
+    supported_lines = [line for line in o.strip().splitlines() if "unsupported" not in line]
+    return len(supported_lines)
```
fwyzard, line 93: This approach should indeed work most of the time.

More in general, we can have a machine where only some of the GPU are supported; in that case counting the number of supported GPUs may not be enough.
A more general approach would be to return the list of supported GPUs, _e.g._ `[0, 1, 3]` instead of their number.

smuzaffar: or may be `{cuda|rocm}ComputeCapabilities` already dump this information by adding `supported | unsupported`  key words?

smuzaffar: > Sure, I can do that.
> 
> But we still need to collect the indices (or ids) of the supported GPUs, we cannot just count how many there are.

sure, if `{cuda|rocm}ComputeCapabilities ` can just dump the gpus ids which we can use then that will help

fwyzard: > or may be `{cuda|rocm}ComputeCapabilities` already dump this information by adding `supported | unsupported` key words?

this is already the case, it prints `(unsupported)` for the unsupported ones.

iarspider: > A more general approach would be to return the list of supported GPUs, _e.g._ `[0, 1, 3]` instead of their number.

Thanks, implemented that.

jobs/jobscheduler.py
```diff
@@ -349,7 +367,7 @@ def initJobs(jobs, resources, otype):
                 if (opts.maxmemory > 0)
                 else int(MachineMemoryGB * 1024 * 1024 * 10.24 * opts.memory)
             ),
-            "gpu": list(range(8)),
+            "gpu": CUDAGPUList + ROCMGPUList,
```
smuzaffar, line 370: @iarspider , I think IB relval job will hang and do nothing it both `CUDAGPUList`  and `ROCMGPUList` are empty  ( may be due to gpu/driver issue the `xxxComputeCapabilities` does not return any thing)

jobs/jobscheduler.py
```diff
@@ -215,8 +231,10 @@ def initJobs(jobs, resources, otype):
             total_jobs += 1
             job = group["commands"][i]
             job["origtime"] = job["time"]
-            if "HIP_VISIBLE_DEVICES=0" in job["command"]:
-                job["gpu"] = -1
+            job["gpu"] = -1
```
smuzaffar, line 234: @iarspider , do not add `gpu` if job command does not need it

jobs/jobscheduler.py
```diff
@@ -70,8 +71,23 @@ def runJob(job):
         job["exit_code"] = 0
     else:
         cmd = job["command"]
-        if ("gpu" in job) and (job["gpu"] >= 0):
-            cmd = cmd.replace("HIP_VISIBLE_DEVICES=0", "HIP_VISIBLE_DEVICES=%s" % job["gpu"])
+        if job.get("gpu", -1) >= 0:
+            maker, idx = job["gpu"].split(":", 1)
```
smuzaffar, line 75: if job["gpu"] is iof form `make:id` then I think the above condition ` job.get("gpu", -1) >= 0` will not work

jobs/jobscheduler.py
smuzaffar, line 370: @iarspider , though for now this is going to work but what will happen if we have a host with both nvidia and amd gpus? On such nodes though cmssw is only going to use either rocm or cuda but job scheduler here will keep on submitting jobs as gpu list here will have combined list of gpus. 

I would suggest to to replace `gpu` with `cuda` and `rocm` and if command has `HIP_VISIBLE_DEVICES=[0-9]+` then use `rocm` and if command has `CUDA_VISIBLE_DEVICES=[0-9]*` then use cuda resources

jobs/jobscheduler.py
```diff
@@ -358,6 +376,8 @@ def initJobs(jobs, resources, otype):
     }
     print(MachineCPUCount, MachineMemoryGB, resources)
     jobs = initJobs(json.load(open(opts.jobs)), resources, opts.type)
+    if any(j.get("gpu") for j in jobs) and not resources["total"]["gpu"]:
+        raise RuntimeError("One or more jobs require GPU, but no usable GPUs found")
```
smuzaffar, line 380: @iarspider , I tried using this branch for one of the relval tests and it failed with error
```
  File "/data/cmsbld/jenkins/workspace/ib-run-relvals/cms-bot/jobs/jobscheduler.py", line 379, in <module>
    if any(j.get("gpu") for j in jobs) and not resources["total"]["gpu"]:
  File "/data/cmsbld/jenkins/workspace/ib-run-relvals/cms-bot/jobs/jobscheduler.py", line 379, in <genexpr>
    if any(j.get("gpu") for j in jobs) and not resources["total"]["gpu"]:
AttributeError: 'str' object has no attribute 'get'
```

jobs/jobscheduler.py
smuzaffar: as discussed our mattermost, the current logic to assigning any available gpu matches what runTheMatrix does. So no need to break `gpu` in to `cuda/rocm`

jobs/jobscheduler.py
smuzaffar: @iarspider , I think this is still not complete. `jobs` has the follow format
```
{
  "env": {}, 
  "final_job": "echo All Done", 
  "final_per_group": {
    "command": "/data/cmsbld/jenkins/workspace/ib-run-relvals/cms-bot/jobs/workflow_final.py %(jobs_results)s", 
    "cpu": 10, 
    "rss": 10485760, 
    "time": 30
  }, 
  "jobs": [
    {
      "commands": [
        {
          "command": "cd 145.405_RunMuon02024F; cmsDriver.py step2  --process reHLT -s L1REPACK:Full,HLT:@relval2025 --conditions auto:run3_hlt_relval --data  --eventcontent FEVTDEBUGHLT --datatier FEVTDEBUGHLT --era Run3_2024  --customise Validation/Performance/TimeMemorySummary.customiseWithTimeMemorySummary --prefix 'python2 /data/cmsbld/jenkins/workspace/ib-run-relvals/cms-bot/monitor_workflow.py timeout --signal SIGTERM 9000 '  -n 100  --filein filelist:step1_dasquery.log --lumiToProcess step1_lumiRanges.log --fileout file:step2.root  --suffix \"-j JobReport2.xml \"  --nThreads 4 > step2_RunMuon02024F.log  2>&1", 
          "cpu": 99, 
          "cpu_avg": 199, 
          "cpu_max": 299, 
          "rss": 112873472, 
          "rss_avg": 2216132608, 
          "rss_max": 4319391744, 
          "time": 560
        }, 
        {
          "command": "cd 145.405_RunMuon02024F; cmsDriver.py step3  --conditions auto:run3_data_prompt_relval -s RAW2DIGI,L1Reco,RECO,PAT,NANO,DQM:@standardDQM+@miniAODDQM+@nanoAODDQM --datatier RECO,MINIAOD,NANOAOD,DQMIO --eventcontent RECO,MINIAOD,NANOEDMAOD,DQM --data  --process reRECO --scenario pp --era Run3 --customise Configuration/DataProcessing/RecoTLR.customisePostEra_Run3 --hltProcess reHLT  --customise Validation/Performance/TimeMemorySummary.customiseWithTimeMemorySummary --prefix 'python2 /data/cmsbld/jenkins/workspace/ib-run-relvals/cms-bot/monitor_workflow.py timeout --signal SIGTERM 9000 '  -n 100  --filein  file:step2.root  --fileout file:step3.root  --suffix \"-j JobReport3.xml \"  --nThreads 4 > step3_RunMuon02024F.log  2>&1", 
          "cpu": 399, 
          "cpu_avg": 408, 
          "cpu_max": 418, 
          "rss": 7662981120, 
          "rss_avg": 7769253888, 
          "rss_max": 7875526656, 
          "time": 339
        }, 
        {
          "command": "cd 145.405_RunMuon02024F; cmsDriver.py step4  -s HARVESTING:@standardDQM+@miniAODDQM+@nanoAODDQM --conditions auto:run3_data --data  --filetype DQM --scenario pp --era Run3_2024  --customise Validation/Performance/TimeMemorySummary.customiseWithTimeMemorySummary --prefix 'python2 /data/cmsbld/jenkins/workspace/ib-run-relvals/cms-bot/monitor_workflow.py timeout --signal SIGTERM 9000 '  -n 100  --filein file:step3_inDQM.root --fileout file:step4.root  --suffix \"-j JobReport4.xml \"  > step4_RunMuon02024F.log  2>&1", 
          "cpu": 99, 
          "cpu_avg": 104, 
          "cpu_max": 109, 
          "rss": 2077976576, 
          "rss_avg": 2118272000, 
          "rss_max": 2158567424, 
          "time": 21
        }
      ], 
      "name": "145.405"
    }, 
    {
      "commands": [
        {
          "command": "cd 11608.0_SingleMuPt100+2022; cmsDriver.py step2  -s DIGI:pdigi_valid,L1,DIGI2RAW,HLT:@relval2022 --conditions auto:phase1_2022_realistic --datatier GEN-SIM-DIGI-RAW -n 10 --eventcontent FEVTDEBUGHLT --geometry DB:Extended --era Run3  --customise Validation/Performance/TimeMemorySummary.customiseWithTimeMemorySummary --prefix 'python2 /data/cmsbld/jenkins/workspace/ib-run-relvals/cms-bot/monitor_workflow.py timeout --signal SIGTERM 9000 '  --filein filelist:step1_dasquery.log --fileout file:step2.root  --suffix \"-j JobReport2.xml \"  --nThreads 4 > step2_SingleMuPt100+2022.log  2>&1", 
          "cpu": 349, 
          "cpu_avg": 379, 
          "cpu_max": 409, 
          "rss": 2581639168, 
          "rss_avg": 2926827520, 
          "rss_max": 3272015872, 
          "time": 25
        }, 
        {
          "command": "cd 11608.0_SingleMuPt100+2022; cmsDriver.py step3  -s RAW2DIGI,L1Reco,RECO,RECOSIM,PAT,NANO,VALIDATION:@standardValidationNoHLT+@miniAODValidation,DQM:@standardDQMFakeHLT+@miniAODDQM+@nanoAODDQM --conditions auto:phase1_2022_realistic --datatier GEN-SIM-RECO,MINIAODSIM,NANOAODSIM,DQMIO -n 10 --eventcontent RECOSIM,MINIAODSIM,NANOEDMAODSIM,DQM --geometry DB:Extended --era Run3  --customise Validation/Performance/TimeMemorySummary.customiseWithTimeMemorySummary --prefix 'python2 /data/cmsbld/jenkins/workspace/ib-run-relvals/cms-bot/monitor_workflow.py timeout --signal SIGTERM 9000 '  --filein  file:step2.root  --fileout file:step3.root  --suffix \"-j JobReport3.xml \"  --nThreads 4 > step3_SingleMuPt100+2022.log  2>&1", 
          "cpu": 99, 
          "cpu_avg": 159, 
          "cpu_max": 219, 
          "rss": 3501420544, 
          "rss_avg": 4861636608, 
          "rss_max": 6221852672, 
          "time": 80
        }, 
...
...
```

so we should be looking for `gpu`  for each command

cmsutils.py
```diff
@@ -70,6 +70,35 @@ def _memorySizeGB():
...
+    supported_lines = [
+        "{0}:{1}".format(gpu_type, i)
+        for i, line in enumerate(o.strip().splitlines())
+        if "unsupported" not in line
+    ]
+    return supported_lines
+
+
+CUDAGPUList = _gpuList("cuda")
+ROCMGPUList = _gpuList("rocm")
```
smuzaffar, line 101: @iarspider , do not create these global variables ( this means every time this file is imported then we run `xxxComputeCapabilities`). Instead change the name of `_gpuList()` and directly call it from jobscheduler.py

jobs/jobscheduler.py
```diff
@@ -358,6 +375,11 @@ def initJobs(jobs, resources, otype):
     }
     print(MachineCPUCount, MachineMemoryGB, resources)
     jobs = initJobs(json.load(open(opts.jobs)), resources, opts.type)
+
+    needGPU = any("_VISIBLE_DEVICES" in c["command"] for j in jobs["jobs"] for c in j["commands"])
```
smuzaffar, line 379: why not just
```
needGPU = any("gpu" in c for j in jobs["jobs"] for c in j["commands"])
```

cmsutils.py
iarspider: Ok

jobs/jobscheduler.py
iarspider: Because there is no `gpu`  in command, e.g. `cd 141.044412_Run3-2023_JetMET2023D_RecoECALOnlyGPU; CUDA_VISIBLE_DEVICES=0 HIP_VISIBLE_DEVICES= cmsDriver.py step2  --process reHLT -s L1REPACK:Full,HLT:@relval2023 --conditions auto:run3_hlt_relval --data  --eventcontent FEVTDEBUGHLT --datatier FEVTDEBUGHLT --era Run3_2023  --customise Validation/Performance/TimeMemorySummary.customiseWithTimeMemorySummary --prefix 'python2 .../monitor_workflow.py timeout --signal SIGTERM 9000 '  -n 100  --filein filelist:step1_dasquery.log --lumiToProcess step1_lumiRanges.log --fileout file:step2.root  --suffix \"-j JobReport2.xml \"  --nThreads 4 > step2_Run3-2023_JetMET2023D_RecoECALOnlyGPU.log  2>&1`

smuzaffar: `j["command"]` is a dict and we add `gpu` in it in the `initJobs` function which was called just above this line

smuzaffar:
I tried running it and it failed with error
```
  File "/data/cmsbld/jenkins/workspace/ib-run-relvals/cms-bot/run-ib-relval.py", line 12, in <module>
    from os import environ, system, waitpid, waitstatus_to_exitcode
ImportError: cannot import name waitstatus_to_exitcode
```

run-ib-relval.py
```diff
@@ -10,6 +10,20 @@
 from sys import exit, argv
 from optparse import OptionParser
 from os import environ, system, waitpid
+
+try:
+    from os import waitstatus_to_exitcode
+except ImportError:
+
+    def waitstatus_to_exitcode(status):
```
smuzaffar, line 18: @iarspider , `os` is not imported so calls like ` os.WIFEXITED` will fail . By the way what benefit we get from using `waitstatus_to_exitcode`  ? Note that exit code here are not from the `runTheMatrix/cmsDriver` commands  so basically we do not care much about the exit code . if `jobscheduler.py` fails then we mark jenkins job failed so no need to use ` waitstatus_to_exitcode`

smuzaffar:
@iarspider , how many more week will it take to get this fixed? This feature should not have take more than an hour to implement

run-ib-relval.py
```diff
@@ -116,7 +116,7 @@ def process_relvals(threads=None, cmssw_version=None, arch=None, cmssw_base=None
             e = waitpid(p.pid, 0)[1]
             print("Time took to create jobs:", int(time() - stime), "sec")
             if e:
-                exit(e)
+                exit(0 if e == 0 else 1)
```
smuzaffar, line 119: @iarspider, sorry I am not able to understand this change.`waitpid(p.pid, 0)[1]` is already return `0` (in case of success) or `non-0`  (in case of failure) so whats wrong with `exit(e)`?

iarspider: Because `e` is not a valid exit code (e.g. it will be 256 if `runTheMatrix` exits after throwing exception), and `exit(256)` is treated as `exit(0)`:

```
$ python3 -c "exit(256)"; echo $?
0
```

iarspider: So `e` is either `-N` if signal N occurs, or `exit_code << 8` otherwise. We either decode it with `waitstatus_to_exitcode`, or just check for zero/non-zero.

smuzaffar: I get e>0 when kill signal is send to the process but yes if `exit(256)` causes python to exit with `0` then  current change is acceptable

jobs/create-relval-jobs.py
```diff
@@ -37,6 +37,7 @@ def createJob(workflow, cmssw_ver, arch):
     e, o = run_cmd(cmd)
     if e:
         print("ERROR:%s:%s" % (workflow, o))
+        exit(1)
```
smuzaffar, line 40: @iarspider , I do not think we want to add this. The command above can fail e.g. it can fail to run das-client for some workflows but we still want to run other workflows. So please remove this

smuzaffar: is `exit` here going to terminate the main python process or just the thread?

smuzaffar: looking at the code in the `try` block, I am sure we do not want to exit. So please remove it

smuzaffar: jobscheduler should never fail because of the commands it runs. commands/relvals failures are reported in the IB dashboard via the workflow.log file we create in the `try` block

smuzaffar: by the way, does runTheMatrix create `runall-report-step123-.log` for the workflow if it is asked to use GPU and n GPU is found?

smuzaffar: ok, we can not just simple exit from here on failure. This code runs the relval step1 (to run the das-client) and create cmdLog. We do not want to exit the relval jenkins jobs as one das-cleint step failed.

I suggest to remove the exit so that we can move forward with this PR. 
Later we can add checks in run-ib-relval.py to process the jobs/commands and fail if needed

iarspider: Sure. For the record: in this state, the Jenkins job will not fail if GPU is required but not found, but will also not run any workflows.

smuzaffar: sure but I think even with `exit(1)` will not fail the job as this only exists the thread `create-relval-jobs.py` will still exit with exit code `0` :-)

smuzaffar:
please test

smuzaffar:
+externals

PR #2471, opened 2025-04-08
Do not process closed PRs

iarspider:
@smuzaffar as discussed

### DIFF
--- a/process_pr.py
+++ b/process_pr.py
@@ -2350,7 +2350,7 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
             add_labels = repo_config.ADD_LABELS
         except:
             pass
-        if add_labels:
+        if add_labels and create_status:
             issue.edit(labels=list(labels))
 
     if not issue.pull_request:
--- a/process_pr.py
+++ b/process_pr.py
@@ -1019,6 +1019,7 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
     ok_too_many_files = False
     warned_too_many_files = False
     is_draft_pr = False
+    create_status = True
 
     if issue.pull_request:
         pr = repo.get_pull(prId)
@@ -2010,6 +2011,9 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
     #    if not extra_labels[ltype]:
     #        del extra_labels[ltype]
 
+    if bot_status is None and issue.state != "open":
+        create_status = False
+
     # Always set test pending label
     if "tests" in signatures:
         if test_comment is not None:
@@ -2102,7 +2106,7 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
                                 "Some test might have been restarted for %s. Resetting the status",
                                 status.context,
                             )
-                            if not dryRun:
+                            if create_status and not dryRun:
                                 last_commit_obj.create_status(
                                     "success",
                                     description="OK",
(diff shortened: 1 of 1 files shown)


process_pr.py
```diff
@@ -1043,6 +1043,9 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
                     )
                     issue.create_comment(msg)
             return
+        if pr.merged and (not force):
+            logger.error("This PR is merged, quitting")
+            return
```
smuzaffar, line 1048: adding this check here means bot will not process any comment ... right? what is we want to assign a label for a merged/closed PR?

iarspider: > adding this check here means bot will not process any comment ... right? 

yes

> what if we want to assign a label for a merged/closed PR?

We can run the bot with force flag (I don't think this will happen often). But I can try blocking less if we want to support this usage scenario.

smuzaffar: we do want to process all comments even if PR is merged/closed. Many L2 signs a PR after it has been merged. so force running is not an option.

I would suggest to disable setting commit status and running PR tests if PR is merged/closed and old commit statuses are empty

process_pr.py
```diff
@@ -2393,7 +2398,7 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
 
     logger.debug("All Parameters: %s", global_test_params)
     # For now, only trigger tests for cms-sw/cmssw and cms-sw/cmsdist
-    if create_test_property:
+    if create_test_property and bot_status is not None and not is_closed_pr:
```
smuzaffar, line 2401: @iarspider , checking for `bot_status` has no use here. It might be empty for first run but on next bot run it will have some statuses . all of your current changes are not needed.  A simple change like 
```
if pr.state == "closed":
  create_test_property=False
``` 
after https://github.com/cms-sw/cms-bot/blob/master/process_pr.py#L1080 line will stop running PR tests for closed PRs. 

I also wanted to avoid creating new statuses if PR is closed and old statuses are empty. It could be that there is bug on github side and they restore the old commit statuses at some point. If we create new statuses then they might not be able to restore them

iarspider: Ah, I see. I will now work on the second part of this PR (not updating statuses).

smuzaffar:
instea dof uses `create_status` flag everywhere, as discussed, please try to override the commit status creation function

iarspider:
please test
to check changes. PR temporary ignored by bot.

smuzaffar:
@iarspider , I have tested this for https://github.com/cms-sw/cmssw/pull/40600 and it worked fine but it did change the label to `test-started` though it properly did not create the property file for test to start. Can we avoid changing test label too for such PRs?

iarspider:
@smuzaffar so if we don't create status, we also don't change labels, right?

smuzaffar:
@iarspider , not exactly. We will still allow to change labels ( e.g. when someone signs a already merged PR) but we should avoid changing test labels for closed PRs

iarspider:
> @iarspider , not exactly. We will still allow to change labels ( e.g. when someone signs a already merged PR) but we should avoid changing test labels for closed PRs

Understood.

smuzaffar:
>  we should avoid changing test labels for closed PRs

@iarspider , Note that we do want bot to update the tests labels for PRs for which release managers has merged while tests are still running. So we should avoid changing test labels for closed PRs which also have no commit statuses .

smuzaffar:
+externals

looks good, tested at https://github.com/cms-sw/cmssw/pull/40598 and https://github.com/cms-sw/cmssw/pull/40599 and bot did not set commit statuses or changed test labels

smuzaffar (approved):

PR #2486, opened 2025-04-23
Add more tests and start measuring coverage

iarspider:
@smuzaffar changes to modify_comment.py are a follow-up of our discussion (the two added functions are not used by process_pr, and don't rely on any process_pr parts). 

I also couldn't come up with a scenario in which bot cache will shrink - we never remove anything from cache, so it can only grow. In theory, if we barely exceed the comment limit and get cache to split between two comments, and then update the cache to compress better we _may_ get it to shrink. 

When tests are ran from github workflow, I don't generate coverage report, this is something we need to discuss later - e.g. we can use a webhook trigger Jenkins job to generate coverage report and host it somewhere on cmssdt, or use external service like codecov.

I have changed how actions are compared (recorded vs performed): b64-encoded data is not compared. It turned out that gzip may produce different compressed bytes for the same input in different zlib versions. Instead, I have hooked `write_bot_cache` function to record (and verify) uncompressed cache.


iarspider:
Latest coverage report (86%): https://iarazumov.web.cern.ch/htmlcov/z_e8be2c0e1675be55_process_pr_py.html . @smuzaffar this is now ready for you to review and merge.

modify_comment.py
```diff
@@ -11,6 +11,34 @@
 setdefaulttimeout(120)
 SCRIPT_DIR = dirname(abspath(sys.argv[0]))
 
+
+def find_last_comment(issue, user, match):
+    last_comment = None
+    for comment in issue.get_comments():
+        if (user != comment.user.login) or (not comment.body):
+            continue
+        if not re.match(
```
smuzaffar, line 20: I guess we need to import `re` in this script ... right?

modify_comment.py
```diff
@@ -11,6 +11,34 @@
...
+            match, comment.body.encode("ascii", "ignore").decode().strip("\n\t\r "), re.MULTILINE
+        ):
+            continue
+        last_comment = comment
+        logger.debug("Matched comment from %s with comment id %s", comment.user.login, comment.id)
```
smuzaffar, line 25: where is `logger` coming from?

smuzaffar: do we need `logger` here?

modify_comment.py
```diff
@@ -11,6 +11,34 @@
...
+def modify_comment(comment, match, replace, dryRun):
+    comment_msg = comment.body.encode("ascii", "ignore").decode() if comment.body else ""
+    if match:
+        new_comment_msg = re.sub(match, replace, comment_msg)
+    else:
+        new_comment_msg = comment_msg + "\n" + replace
+    if new_comment_msg != comment_msg:
+        if not dryRun:
+            comment.edit(new_comment_msg)
+            logger.info("Message updated")
```
smuzaffar, line 38: same here ... may be remove logger

process_pr.py
```diff
@@ -1049,7 +1093,8 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
             if pr.state != "closed":
                 logger.error("This pull request must go in to master branch")
                 if not dryRun:
-                    edit_pr(repo.full_name, prId, base="master")
+                    # edit_pr(repo.full_name, prId, base="master")
```
smuzaffar, line 1096: @iarspider , so we can remove the import of `edit_pr` ... right? 
also remove the commented line

smuzaffar:
@iarspider , overall looks good ( just take care of `edit_pr` comemnt).

iarspider:
Ok. I will merge this PR once tests pass.

PR #2488, opened 2025-04-28
Fix signing labels for orp-only issues

iarspider:
If `orp` and/or `operations` are the only assigned categories, do not add `fully-signed` label until they sign the issue (or PR).  This fixes the invalid behavior seen in https://github.com/cms-sw/cmssw/issues/47964

### DIFF
--- a/process_pr.py
+++ b/process_pr.py
@@ -2280,7 +2280,6 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
         x
         for x in labels
         if not x.endswith("-approved")
-        and not x.startswith("orp")
         and not x.startswith("tests")
         and not x.startswith("pending-assignment")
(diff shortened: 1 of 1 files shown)


process_pr.py
```diff
@@ -2280,7 +2280,6 @@ def process_pr(repo_config, gh, repo, issue, dryRun, cmsbuild_user=None, force=F
         x
         for x in labels
         if not x.endswith("-approved")
-        and not x.startswith("orp")
```
smuzaffar, line 2283: @iarspider , isn't `and (not x.startwith("orp") or not issue.pull_request)` enough for this fix?

iarspider: Technically yes. I can change L2292 to check for `not issue.pull_request` instead of checking `signing_categories`, but would prefer to keep that condition separate, since it's more readable I think.

PR #2492, opened 2025-05-06
Remove Sandra from the AlCaDB signatures

perrotta:
Remove Sandra from the AlCaDB signatures because she's not AlCaDB L2 any more since Apr 17


smuzaffar:
+externals

PR #2480, opened 2025-04-17
Add extra logging to try and debug ib-create-tag.py failing 

iarspider:
… (due to HTTP error 422).

PR #2493, opened 2025-05-06
ib-create-tag: do not fail if trying to create a tag that already exists

iarspider:
https://github.com/cms-sw/cms-bot/pull/2480 revealed that we are trying to create same reference twice, [e.g.](https://cmssdt.cern.ch/jenkins/job/ib-tag-and-schdule/9355/consoleFull):

```
create_git_tag(cms-sw/cmssw, CMSSW_15_0_X_2025-04-29-2300, 001b97d2ac867b36ee0fb0273dda20fe6e95f6ed) failed: HTTP Error 422: Unprocessable Entity
Response: {"message":"Reference already exists","documentation_url":"https://docs.github.com/rest/git/refs#create-a-reference","status":"422"}'
```

I think what happens is Github returns 5xx error ("server error"), but since we assume the only error we can get is 404 (i.e. tag doesn't exist) we try to create the reference anyway and fail. 

[Documentation](https://docs.github.com/en/rest/git/refs?apiVersion=2022-11-28#get-a-reference) also mentions code `409 Conflict` (probably when `ref` parameter is not `heads/xyz` or `tags/xyz`).

Should we silently retry a few (5?) times if we get error 5xx when checking for tag existence?

### DIFF
--- a/ib-create-tag.py
+++ b/ib-create-tag.py
@@ -88,7 +88,14 @@ if __name__ == "__main__":
     try:
         ref = get_git_tag(repo, RELEASE_NAME)
         HEAD_SHA = ref["object"]["sha"]
-    except HTTPError:
+    except HTTPError as e1:
+        if e1.code != 404:
+            error_body = e1.read().decode("ascii", errors="replace")
+            print(
+                "Unexpected HTTPError: {0}\nResponse: {1}".format(e1, error_body), file=sys.stderr
+            )
+            sys.exit(1)
+
         commits_ = get_commits(repo, RELEASE_BRANCH, until=ib_date, per_page=100)
         if not commits_:
             sys.exit(1)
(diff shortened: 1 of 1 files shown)


ib-create-tag.py
```diff
@@ -85,43 +88,69 @@ def currenttz():
...
+                    except HTTPError as e:
+                        error_body = e.read().decode("ascii", errors="replace")
+                        if not (e.code == 422 and "already exists" in error_body.lower()):
+                            print(
+                                "Unexpected HTTPError when creating git reference: {0}\nResponse: {1}".format(
+                                    e, error_body
+                                ),
+                                file=sys.stderr,
+                            )
+                            sys.exit(1)
```
iarspider, line 139: Alternatively we can just retry here and not fail

iarspider:
@smuzaffar I would like to merge this tomorrow before 11h IB.

smuzaffar:
just update the jenkins job to use this branch. If things goes fine then we can merge it

iarspider:
ok

PR #2508, opened 2025-05-20
Rpm db size issue

smuzaffar:
work around for installing packages when RPM database is over 2GB ( see https://cern.service-now.com/service-portal?id=ticket&table=incident&n=INC4478007 for details). This is a generic issue with cvmfs and newer linux kernel. Opening a large file with read/write permission fails. A simple `open("/cvmfs/repo/some/large/file",O_RDWR);` fails on kernel  `5.14.0-503` ( which is installed on our cms-ib.cern.ch cvmfs publishers) while it works on kernel `5.14.0-362` ( which is available on cms.cern.ch cvmfs publisher).

The change here moves the large `Packages` file to local disk, create a symlink on cvmfs and then runs `rpm -ivh` and finally move back the `Packages` file to cvmfs. We can remove this workaround once cvmfs or kernel has a fix for it.

### DIFF
--- a/cvmfsInstall.sh
+++ b/cvmfsInstall.sh
@@ -9,7 +9,7 @@ function install_package() {
       echo "  Trying local DB"
       rm -rf ${WORKSPACE}/rpm
       mv $WORKDIR/${SCRAM_ARCH}/var/lib/rpm  ${WORKSPACE}/rpm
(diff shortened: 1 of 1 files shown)


smuzaffar:
+externals 

this has been tested for installing [CMSSW IBs](https://cmssdt.cern.ch/jenkins/job/ib-install-cvmfs/98439/console) for last week for el8_amd64_gcc12 ( for which we had built too many IBs and RPM DB was over 2GB).

PR #2509, opened 2025-05-20
Revert "Remove MKLDNN0 build which is no longer needed"

gartung:
Reverts cms-sw/cms-bot#2503
Turns out an IB with Tensorflow built with OneDNN disabled is still required for profiling to run without segfault.


gartung:
please test

PR #2510, opened 2025-05-20
Debug why bot is setting ack status too many times

iarspider:


iarspider:
@smuzaffar do you mind if I switch cms-bot job to use this branch? PR https://github.com/cms-sw/cmsdist/pull/9845 might be a good way to reproduce the situation where bot sets `ack` status too many times...

iarspider:
> we can but I doubt it this will catch any thing ... feel free to update the job to use this branch. Note that if we push other changes to master branch then we also have to make sure that cms-bot job picks up those changes + this branch changes

Done.

PR #2528, opened 2025-07-09
Disable aarch64 Ibs for CMSSW 14.2 and early release cycles

smuzaffar:
As discussed in Core SW meeting on 8th July 2025, this pull request disables aarch64 IBs for production release cycles earlier 14.2.X and earlier. It also disable el9_aarch64_gcc12 for 15.0.X . All these release cycles use root version 6.32 or earlier.


smuzaffar:
+externals

PR #2533, opened 2025-07-16
Do not retrigger bot on closed PRs

iarspider:


process_pr.py
```diff
@@ -1090,7 +1090,7 @@ def process_pr(
         if (
             repo.full_name == "cms-sw/cms-bot"
             and os.getenv("CMS_BOT_TEST_BRANCH", "master") == "master"
-            and False
+            and pr.state != "closed"
```
smuzaffar, line 1093: change looks reasonable. I am not sure if check for "pr.state == "open"` is better/safer

smuzaffar: that I know, just wondering if there was any reason to use `!=` instead of `==` ( just for my education :-) ) ?

PR #2542, opened 2025-07-30
Add HGCal Raw-data L3 to watchers.yaml

IzaakWN:
Adding myself (@IzaakWN) to packages relevant for development in the HGCal Raw-data group, and adding the same packages to my co-convener Pedro (@pfs).

@pfs, can you confirm, please?


smuzaffar:
+externals

PR #2548, opened 2025-08-13
pileup mapping script added

emcgrady:
Added pullPileup.py which enumerates the all pileup containers known to msPileup. A set of files containing the lfns associated with each container. A space delimited text file is a generated which maps container names to each lfn list.


smuzaffar:
@kpedro88 , yes it should be better to move this script under `rucio` sub-directory. I will do it after merging it

PR #2553, opened 2025-09-01
Added extra nvidia gpus h100 and l40s

smuzaffar:


smuzaffar:
please test

smuzaffar:
please test

run tests for all default gpus

fwyzard:
@smuzaffar looks good to me.

The matrix tests look OK. I only checked some of the logs, and they find the GPUs correctly and run successfully. 

The unit tests are there only for some generic, non-GPU tests, which are skipped in the GPU tests, so there's nothing to see there. 

The addon tests are not run for the GPU case ?
It would actually be useful to run at least the HLT ones, can they be added ?

smuzaffar:
+externals

fwyzard:
> It would be easier for us to integrate this if https://github.com/cms-sw/cmssw/blob/master/Utilities/ReleaseScripts/scripts/addOnTests.py has `--gpu` option which selected some of these to be run

I'll look into it.

PR #2556, opened 2025-09-01
Update generator list

lviliani:
Added Konstantinos Theofilatos (new GEN L2) and Sercan Sen (new GEN Core Integration L3) to `generators`.

The old members have been removed.


smuzaffar:
+externals

changes suggested by @cms-sw/generators-l2

PR #2567, opened 2025-09-14
Fail the run-pr-external_checks if it can not find cmssw-tool-conf

smuzaffar:
`run-pr-external_checks.sh` is silently failing since we move to `cmssw-tools` instead of `cmssw-tool-conf`.  This change makes sure that external checks are marked failed if they can not find cmssw-tool-conf.


smuzaffar:
please test with cms-sw/cmsdist#10074

with current change the external checks should fail as they will not be able to find cmssw-tools

smuzaffar:
please test with cms-sw/cmsdist#10072

PR #2580, opened 2025-09-30
Allow marking categories deprectated

iarspider:


iarspider:
This will block `assign upgrade`, but not auto-assignment from file names. Right now I don't see a clean way of doing it (except adding a date to `legacy_categories` entries).

After this PR is merged, we can remove all packages and L1s from `upgrade` category, but keep the category itself.

iarspider:
I am considering a different approach: a category has a timestamp, and any PR/issue created after that time will ignore that category: i.e. will not assign from `assign` command or from changed files. @smuzaffar does this sound like a reasonable approach?

process_pr.py
```diff
@@ -1022,13 +1029,29 @@ def process_pr(
...
+    CMSSW_CATEGORIES = copy.deepcopy(default_CMSSW_CATEGORIES)
+
+    # LEGACY_CATEGORIES is a mapping from name to datetime (when the category becomes legacy)
+    # Extract categories that are legacy at the moment of issue creation
+    lc = getattr(repo_config, "LEGACY_CATEGORIES", {})
+    issue_created_at: datetime = issue.created_at
+    if issue_created_at.tzinfo is None:
+        issue_created_at = issue_created_at.replace(tzinfo=timezone.utc)
+
+    for lc in (cat for cat, ts in lc.items() if issue_created_at > ts):
```
smuzaffar, line 1043: @iarspider , may be change either `lc` loop variable name  here or rename `lc` variable on line 1038.

iarspider: Done

smuzaffar:
looks good to me, 
@srimanob , this change should allow us to keep the already assigned `ugrade` labels for old PRs and avoid assignng it new PRs. If you have no objections then we can move forward with integrating it (may be changing the `ugrade` deprecated  time to future e.g. 3rd of OCT)

iarspider:
the test failure (order in which packages are listed in a comment is different) is unrelated, I would prefer to fix it in a separate PR.

PR #2581, opened 2025-10-01
Fix wrong name in the generators category

lviliani:
We were using the wrong github account for Sercan Sen (GEN Core Software Integration convener).
This PR is fixing the issue.


iarspider:
+externals

PR #2583, opened 2025-10-03
fix-igprof-sql.py: Use python3. Use open with encoding="latin-1", errors="ignore"

gartung:


gartung:
please test

gartung:
please test

smuzaffar:
@gartung , using `python3` is safe as we only run igprof for 14.0.X and above so python3 should be available. Did you check if double quote change will work for old sqlite version?

gartung:
Accepting double quotes for string literals was disabled for the sqlite3 in AL9.x

PR #2588, opened 2025-10-07
added script to convert releases.map in to json format with ordered list of releases

smuzaffar:
See https://github.com/cms-sw/cmssw/issues/48837 for details


smuzaffar:
lets get this merged and update jenkins job to run it so that we get releases.json automatically generated and deployed on cmssdt and cvmfs

PR #2596, opened 2025-10-16
Make categories and labels per-repository

iarspider:


iarspider:
@smuzaffar I suggest we revert and redo this: we need to separate manual assignment of categories (`assign externals`) from automatic one (via changed files).

PR #2609, opened 2025-11-05
Replace BTV-POG L2s with ones presently in charge

battibass:
PR made upon request of @acalandr and @AlexDeMoor.


mandrenguyen:
+1

PR #2612, opened 2025-11-06
Avoid leaking file descriptors when reading GPU lists

iarspider:
Technical PR. In principle the descriptor will get released once the list comprehension `[...]` is complete [a], but it's better to be explicit

[a]
Here’s what actually happens:

1. `open(...)` creates a file object.
2. `.read()` reads the **entire contents** of the file into memory as a single string.
3. After `.read()` finishes, **the file object still exists**, but you’ve lost the reference to it (since you never assigned it to a variable).
4. Because you lost the reference, the file object becomes *unreachable* once the expression completes, and Python’s **garbage collector** will eventually close it *when the object is destroyed*.

However — and this is key — **there’s no guarantee *when* that destruction happens**.

In CPython (the most common implementation), it’ll usually happen immediately (due to reference counting).


iarspider:
unhold

smuzaffar:
+externals

PR #2616, opened 2025-11-12
Reco Event Loop will link to compressed Vtune gprof-cc text report.

gartung:
@smuzaffar


smuzaffar:
@gartung , this is now merged and deployed. Now clicking on the `Reco Event loop` link downloads the gz file. Is this what you  expected ?

gartung:
Yes for now. I am working on scripts to produce sqlite3 database and cgi script to display it like igprof-navigator.

PR #2621, opened 2025-11-19
fix release queue regexp to allow using an exact IB for PR tests

smuzaffar:


smuzaffar:
+externals
ci test passed for process_pr

PR #2624, opened 2025-11-25
Add `PhysicsTools/{PyTorchAlpaka,PyTorchAlpakaTest}` for heterogeneous and ml

makortel:
Following https://github.com/cms-sw/cmssw/pull/47984#issuecomment-3577000277


fwyzard:
+1

thanks @makortel

PR #2625, opened 2025-11-26
Add CPP23 IB

iarspider:
Note: we do not want to enable it right away, without making sure all the externals build fine.


iarspider:
@smuzaffar local build  of cmssw-tool-conf succeeded.

PR #2644, opened 2025-12-22
Increase MaxMemoryPreload error threshold to 90 MB

makortel:
We see random variations up to 83 MB (that need to be understood separately, see https://github.com/cms-sw/cmssw/issues/46966), so this increase should reduce noise.

Resolves https://github.com/cms-sw/framework-team/issues/1741


smuzaffar:
+externals

PR #2649, opened 2026-01-11
Added RNTuple PR tests

smuzaffar:


smuzaffar:
please test

smuzaffar:
please test

smuzaffar:
please test with https://github.com/cms-sw/cmssw/pull/49674

makortel:
> Currently for `rntuple` relvals bot does not run comparison job as that needs more changes (comparing baseline root files with PR rntuple root files)

We can live with that for now.

Actually, the DQM output files are still `.root` and should be technically directly usable to compare with the baseline `.root` files. The validateJR ("reco comparisons") would need a complete overhaul for RNTuple that, I think, should be followed up separately (starting with the question "who should do that").

PR #2651, opened 2026-01-14
enable dqm comparison job for rntuple

smuzaffar:
This change enables running PR DQM/Reco comparison for RNTuple tests.

PR #2654, opened 2026-01-16
Report workflows failing maxmem check as a numericly sorted list

gartung:
Tested locally with sample output and produces expected output without python failures.


smuzaffar:
+externals

looks good, tested via https://github.com/cms-sw/cms-bot/pull/2651

Issue #2658, opened 2026-01-20
getting logs out of hlt timing tests


mmusich:
Currently in PR tests we only get the output chart, see e.g. for example [here](https://github.com/cms-sw/cmssw/pull/49821#issuecomment-3761998681).
I wonder if we could also publish the full log of the test, i.e. this file `hlt-p2-timing.log` here:

https://github.com/cms-sw/cms-bot/blob/1bd57f329468d112a9dce9e52060020f22d32ad9/pr_testing/run-pr-hlt-p2-timing.sh#L20

this would allow to shed some light about e.g. the throughput of the timing measurement.

@VourMa @rovere

PR #2662, opened 2026-01-21
P2 HLT timing: check on all expected json files to declare success

mmusich:
Fixes https://github.com/cms-sw/cms-bot/issues/2658.
Title says it all, we need to check all (release-dependent) flavours after PR https://github.com/cms-sw/cmssw/pull/49279

### DIFF
--- a/pr_testing/run-pr-hlt-p2-timing.sh
+++ b/pr_testing/run-pr-hlt-p2-timing.sh
@@ -51,6 +51,7 @@ done
 if [ $missing -eq 0 ]; then
   CHART_URL="https://cmssdt.cern.ch/circles/web/piechart.php?data_name=hlt-p2-timing&resource=time_thread&filter=${CMSSW_VERSION}&dataset=${UPLOAD_PATH}/Phase2Timing_resources"
   echo "HLT_P2_TIMING;SUCCESS,HLT Phase 2 timing Test,See Chart,${CHART_URL}" >> ${RESULTS_DIR}/hlt-p2-timing.txt
+  echo "HLT_P2_TIMING_LOG;OK,HLT Phase 2 timing Test Log,See Logs,hlt-p2-timing.log" >> ${RESULTS_DIR}/hlt-p2-timing.txt
   echo -e "**HLT P2 Timing**: [chart](${CHART_URL})" > ${RESULTS_DIR}/11-hlt-p2-timing-report.res
 
   mv $WORKSPACE/rundir/Phase2Timing*.json $WORKSPACE/json_upload


pr_testing/run-pr-hlt-p2-timing.sh
```diff
@@ -23,7 +23,22 @@ popd
 # Upload results
 source $WORKSPACE/cms-bot/jenkins-artifacts
 touch ${RESULTS_DIR}/11-hlt-p2-timing-report.res ${RESULTS_DIR}/11-hlt-p2-timing-failed.res
-if [ -f $WORKSPACE/rundir/Phase2Timing_resources.json ] ; then
+
+required_files=(
+    "$WORKSPACE/rundir/Phase2Timing_resources.json"
+    "$WORKSPACE/rundir/Phase2Timing_OnCPU_resources.json"
+    "$WORKSPACE/rundir/Phase2Timing_resources_NGT.json"
+)
```
smuzaffar: @mmusich , do old releases also generate these files? Note that we are running these tests for CMSSW_14_1_X and above.

For IBs tests, we are uploading `Phase2Timing_resources*.json` files , should we change it to upload `mv Phase2Timing_*.json` ?

mmusich: @smuzaffar, thank you for reviewing.

>  do old releases also generate these files? Note that we are running these tests for CMSSW_14_1_X and above.

This is the current situation:
- CMSSW_16_1_X: generates all three files
- CMSSW_16_0_X: generates all three files
- CMSSW_15_1_X: generates `Phase2Timing_resources.json` and `Phase2Timing_resources_NGT.json`
- CMSSW_15_0_X (and below): generates `Phase2Timing_resources.json`

how do you suggest to support a different number of files? I'd like to be fully explicit on the expected name of the json file: do we have means from the bash script to know the release version?

> For IBs tests, we are uploading `Phase2Timing_resources*.json` files , should we change it to upload mv `Phase2Timing_*.json` ?

No. Actually thanks for pointing this out. This explains the mystery at https://github.com/cms-sw/cmssw/pull/49279#issuecomment-3484469099 !
I will rather change the name of the output json in order to match the expected pattern.
By the way can you show me which file is responsible for generating the IB tests?

Thanks!

smuzaffar:
@mmusich , p2 hlt node ( whitch nvidia T4 GPU) looks fine now. In past 18 hours, it has successfully ran all IBs unittest, relvals [jobs](https://cmssdt.cern.ch/jenkins/computer/cms%2Dhlt%2D01/builds) for 14.1 and above release cycles. I also that it managed to run PR tests for 16.0.X. 

could it be that issue is only with cms-sw/cmssw#49932 ? may be we can restart PR tests for cms-sw/cmssw#49932 to see if that breaks the node or not?

mmusich:
> Let's get CMSSW_16_0_0 out before breaking the node again ?

agreed.

> could it be that issue is only with https://github.com/cms-sw/cmssw/pull/49932 ? 

It's possbile.

> may be we can restart PR tests for https://github.com/cms-sw/cmssw/pull/49932 to see if that breaks the node or not?

not now (also I am waiting for the timing results to come out of https://github.com/cms-sw/cmssw/pull/49637, so let's wait for that and 16_0_0 to come out).

smuzaffar:
https://github.com/cms-sw/cms-bot/pull/2667 disabled T4 GPU tests but p2 hlt timing tests can still run if explicitly asked via PR comment

PR #2665, opened 2026-01-27
make `PR_TEST_MATRIX_EXTRAS_GPU` use 2026-based relvals instead of 2025

mmusich:
In PR:
  * https://github.com/cms-sw/cmssw/pull/49747 (16.1.X)
  * https://github.com/cms-sw/cmssw/pull/49766 (16.0.X)
  
we changed the default HLT menu run in 2025 relvals to be the `Fake` one.
This makes the relval GPU matrix not as useful as it could be, because we can't run the heterogenous reco in the HLT itself.
Thsi PR proposes to run the 2026 equivalent workflows that are already in the relval_gpu matrix.

@AdrianoDee FYI


mmusich:
> -1

these failures are unrelated, due to https://github.com/cms-sw/cmssw/issues/49795

PR #2674, opened 2026-02-05
Remove unused RUNALLSTEPS variable and the call to runall_cpu.sh in PR profiling script

gartung:
Removes run of Igprof -pp which is not needed anymore.


smuzaffar:
+externals

PR #2681, opened 2026-02-11
pass extra options to das_cache.py for proper caching of das results

smuzaffar:
This change allow to capture extra options of dasgo client e.g. `--limit N` and/or `-unique` can properly record it for das caching job so that same option can be used to cache the results


smuzaffar:
please test with https://github.com/cms-sw/cmssw/pull/50118

PR #2682, opened 2026-02-11
PR profiling: remove check on allowed workflows = default workflows

gartung:


gartung:
please test

makortel:
> I see the concept of "allowed profiling workflows" was added in #1545

From the discussion in that PR it is not clear to me why arbitrary workflows would not be allowed to be run through profiling.

makortel:
In the case profiling workflow(s) are specified via the test parameters, I'd think it would be sufficient to run only those workflows (even if the general philosophy is to specify additional workflows) if that would save resources.

gartung:
please test

gartung:
please test

gartung:
please test

makortel:
@smuzaffar Do you remember why in the past (e.g. in https://github.com/cms-sw/cms-bot/pull/1545) we only allowed specific workflows to be used in `profiling` test?

smuzaffar:
> @smuzaffar Do you remember why in the past (e.g. in #1545) we only allowed specific workflows to be used in `profiling` test?

@makortel , do not remember ? Could it be that for profiling comparison we were only allowing to run the works which we might have run for IBs?

gartung:
@smuzaffar How can I make this schedule the profiling for the IB on demand for workflow(s) specified in pull request and then wait before having the profiling for the IB + PR done?

gartung:
I updated the `ib-run-pr-propfiling` job to run a blocking call to `ib-run-profiling` with the same parameters.

gartung:
please test

gartung:
No. I still need to work out to run with the IB alone "on-demand". I may just add that to the PR profiling script.

PR #2684, opened 2026-02-13
Replace profiling workflow 13034.21 with 18634.21

gartung:


gartung:
I see that 18634.0 is run is IB's. It's just not a "ProdLike" workflow.

gartung:
For the 29834.21 workflow the diff table is available
https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-7f8a85/51320/profiling/index-29834.21.html

PR #2686, opened 2026-02-18
Allow setting profiling workflow by release

gartung:


gartung:
The drop down menu from clicking on the green check mark shows that both workflows were run but only the 17034.21 workflow was reported in the summary. I think I know what script to check.

gartung:
please test for CMSSW_16_0_X

mandrenguyen:
+1
@cms-sw/externals-l2 good for you?

smuzaffar:
@gartung , I do not see profiling results for the last run

PR #2696, opened 2026-03-05
Add beginRun and beginLuminosityBlock to moduleAllocMonitors diff table.

gartung:


makortel:
Looks good for this iteration. I have some ideas for further improvements, but I have a feeling they would require some interactivity with e.g. JavaScript, and thus I'd like to discuss those first.

PR #2698, opened 2026-03-06
Present profiling summary as a table of links with the x axis be the step and the y axis the test name

gartung:


gartung:
test parameters:

enable = profiling
workflows_profiling = 18634.21

PR #2717, opened 2026-04-08
ModuleAllocMonitor diff table updates

gartung:
Makes the ModlueAllocMonitor diff table display only the column of the metric of interest for IB, PR or PR-IB.  
Added search box for  label, type and/or record and displays only the rows matching.
Adds a button to hide rows with a zero metric.


gartung:
test parameters:

workflows_profiling = 18634.0
enable = profiling

gartung:
please test

gartung:
@makortel

smuzaffar:
@gartung , is this ready to go in?

makortel:
Not yet

gartung:
please test

makortel:
Thanks for the update. A few comments based on https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-bf579c/52617/profiling/18634.0/diff-step3_moduleAllocMonitor.circles.json.html

image

The units (`(kB)`) should be moved to where the metric is selected (a dropdown menu?)

Sorting seems work weirdly. If I click once in the `Metric PR-IB` cell, I get
image
where the top values in that column are now smaller than in the first screenshot.

Clicking a second time in the `Metric PR-IB` cell, I get
image
where the values of `0.0` could ok (except there are rows with negative values). But that screenshot shows rows with negative values for `Metric IB` and `Metric PR`, which seems weird (although I guess could theoretically be possible).

The EventSetup modules show a few weirdness more
image

has two `Sum All` rows. This one shows even more `Sum All` rows
image

gartung:
This change was just to get the table to look like what your mockup. 
The action of the row sort is still a work in progress.
For each name/type/record triplet are the rows for Construction and Sum All with the corresponding construction and added total metrics, and Record: record-name with the event setup metrics. That is why multiple Construction and Sum ALL metrics are listed.

gartung:
Will resubmit once https://github.com/cms-sw/cms-bot/pull/2721 is merged since both pull requests change the same file.

PR #2719, opened 2026-04-10
Disable JR comparison tests for EVOLUTION_X PRs

smuzaffar:
@makortel , this PR proposes to disable JR comparison jobs for EVOLUTION_X


smuzaffar:
please test

makortel:
Does this disable also DQM comparisons? Those should be technically doable (as the relevant ROOT files contain only histograms), and could be useful to verify that the the monitored quantities are not changed under these modifications.

smuzaffar:
yes this disables all the comparison tests. Do you only want to disable JR-Comparison?

makortel:
> Do you only want to disable JR-Comparison?

If that is easily doable, yes, I'd like to disable only the JR comparison.

smuzaffar:
I have updated the PR to only disable JR comparison for EVO PRs tests

smuzaffar:
please test for CMSSW_17_0_EVOLUTION_X

makortel:
Thanks!

PR #2721, opened 2026-04-14
ModuleAllocMonitor diff table: If the module type/label/record triplet is missing from the PR or IB data show the missing numbers as N/A; add more transitions.

### LINKED ISSUE #2102: Add more quantities to the ModuleAllocMonitor diff table
### Description

Add all transitions that are in the circles.json file. Make the quantity (e.g. added, nAlloc) selectable e.g. via a dropdown menu.

### LINKED ISSUE #2113: Make ModuleAllocMonitor diff tables to work properly if a module doesn't exist in baseline or in PR
### Description

See https://github.com/cms-sw/cmssw/pull/50292#issuecomment-4143122014

### LINKED ISSUE #2165: Add more transitions to the ModuleAllocMonitor diff table
### Description

Continuation of https://github.com/cms-sw/framework-team/issues/2102 to add also all transitions in the circles.json file.

gartung:
Removed module zmumugammaAnalysis from IB data and module zmumugammaOldAnalysis from PR data and verified that missing data and diff data is show as nan in diff table.

image

### DIFF
--- a/comparisons/moduleAllocMonitor-circles-diff.py
+++ b/comparisons/moduleAllocMonitor-circles-diff.py
@@ -24,7 +24,13 @@ NALLOC_BEGIN_LUMI_KEYS = [
     "nAlloc global begin luminosity block",
     "nAlloc stream begin luminosity block",
 ]
-NALLOC_TOTAL_KEYS = ["nAlloc construction", "nAlloc event", "nAlloc event setup"]
+NALLOC_TOTAL_KEYS = [
+    "nAlloc construction",
+    "nAlloc event",
+    "nAlloc event setup",
+    *NALLOC_BEGIN_RUN_KEYS,
+    *NALLOC_BEGIN_LUMI_KEYS,
+]
 
 
 def module_key(module):
--- a/comparisons/moduleAllocMonitor-circles-diff.py
+++ b/comparisons/moduleAllocMonitor-circles-diff.py
@@ -42,7 +42,9 @@ def sum_numeric_values(data, keys, default="N/A"):
 
 
 def sum_with_suffix(data, metric_keys, suffix, default="N/A"):
-    return sum_numeric_values(data, ["%s %s" % (metric, suffix) for metric in metric_keys], default)
+    return sum_numeric_values(
+        data, ["%s %s" % (metric, suffix) for metric in metric_keys], default
+    )
 
 
 def format_metric(value):
@@ -51,7 +53,8 @@ def format_metric(value):
 
 def append_triplet_cell(summary_lines, ib, pr, diff, attrs='align="right"'):
     summary_lines.append(
-        '<td %s>%s<br>%s<br>%s</td>' % (attrs, format_metric(ib), format_metric(pr), format_metric(diff))
+        "<td %s>%s<br>%s<br>%s</td>"
+        % (attrs, format_metric(ib), format_metric(pr), format_metric(diff))
     )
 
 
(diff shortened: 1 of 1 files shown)


gartung:
test parameters:

workflows_profiling = 18634.0
enable = profiling

gartung:
please test

gartung:
@makortel

makortel:
Thanks @gartung. I'm not sure if `nan` is the best way to represent it though, because for the `PR - IB` difference it would be useful to have the number as `0 - <IB number>` if the PR removed a module, or `<PR number> - 0` if the PR added a module. Also `nan` could be easily interpreted as "something went wrong".

How about using `N/A` for the PR/IB value when a module is missing from PR/IB, but calculate the difference as if the PR/IB value was 0?

gartung:
The number are being printed as `%0.2f` so I ended up using `float(nan)`. If text is used then the format string needs to use `%s`. Should I implement it this way?

makortel:
> The number are being printed as `%0.2f` so I ended up using `float(nan)`. If text is used then the format string needs to use `%s`. Should I implement it this way?

How about using `%s` in the cell itself, and add a helper function that formats `str` as is, and everything else as `%.2f` (or pass the float format as an argument for the `%.0f` that is used in the `nAlloc` counts)?

gartung:
image
image

comparisons/moduleAllocMonitor-circles-diff.py
```diff
@@ -252,164 +248,592 @@ def diff_from(metrics, data, data_total, dest, dest_total, res):
...
+            and isinstance(moduleres.get("added construction diff", 0), (int, float))
+            and isinstance(moduleres.get("added global begin run diff", 0), (int, float))
+            and isinstance(moduleres.get("added stream begin run diff", 0), (int, float))
+            and isinstance(
+                moduleres.get("added global begin luminosity block diff", 0), (int, float)
+            )
+            and isinstance(
+                moduleres.get("added stream begin luminosity block diff", 0), (int, float)
+            )
+            else "N/A"
```
makortel: There is a lot of repetition here. First the whole pattern repeats three times (PR, IB, difference), and inside of each assignment the keys are repeated as well. I'd suggest to abstract these avoid repetition.

I see also other places where the very same code repeats three times for PR, IB, and difference.

makortel (commented):
The result looks good. I think it would be useful to reduce repetition over PR, IB, and diff cases in the code.

comparisons/moduleAllocMonitor-circles-diff.py
```diff
@@ -4,28 +4,355 @@
...
+    "nAlloc global begin luminosity block",
+    "nAlloc stream begin luminosity block",
+]
+NALLOC_TOTAL_KEYS = [
+    "nAlloc construction",
+    "nAlloc event",
+    "nAlloc event setup",
+    *NALLOC_BEGIN_RUN_KEYS,
+    *NALLOC_BEGIN_LUMI_KEYS,
+]
```
makortel: Eventually I'd want these to be generalized for other quantities than `added` and `nAlloc` (I see I don't have a task issue about that yet). It might be better to be done in a later PR though.

makortel:
Looks ok to me

gartung:
@smuzaffar

gartung:
@makortel I added a commit to generalize the functions and added the missing metric columns for `nDealloc` `maxTemp` , and `max1Alloc`.

makortel:
The added quantities make the table quite wide, but given the parallel work (https://github.com/cms-sw/framework-team/issues/2102) to show only one (selectable) quantity at a time, I'd be ok with that.

Eventually it would be nice to have the totals summary table to have all these quantities as well. I think that table should be made to have a separate row for each quantity to avoid very wide tables. I'd be fine with postponing that work to subsequent PR, or doing it here.

gartung:
please test

smuzaffar:
@gartung , ping me or @akritkbehera to merge this when ready

gartung:
@smuzaffar @akritkbehera It is ready to merge. The profiling job keeps timing out for reasons unrelated to this PR.

PR #2723, opened 2026-04-17
Remove git clone of patatrack-scripts

smuzaffar:
Removed cloning of patatrack-scripts repository to streamline the script.


smuzaffar:
python module `imp` fix is now merged in to upstream ` patatrack-scripts ` repo , so lets revert the change and start using upstream  patatrack-scripts  repo for HLT P2 timing tests

mmusich:
> so lets revert the change and start using upstream patatrack-scripts repo for HLT P2 timing tests

can I try testing again https://github.com/cms-sw/cmssw/pull/50479 right away?

PR #2727, opened 2026-04-20
Module Alloc Monitor diff table page update

### LINKED ISSUE #2103: Make columns in ModuleAllocMonitor diff table sortable
### Description

So that a user can select which column to use in the sorting, and whether to sort by ascending or descending order.

Sorting should be done by the "PR - IB diff" value in the cells. Although if it would be possible for a user to select whether to use the IB, PR, or diff, that would be awesaome.

gartung:
Use web page with javascript and json diff data to display diff table in a sortable table that is performant.
Web page template created with CoPilot in VSCode.


makortel:
Thanks. Generally the direction looks good. Polishing-level comments
* The `added`, `max1Alloc`, and `maxTemp` quantities would need units (which is kB, right?)
  * I'd put those at least in the drop-down
  * Column headings would be nice as well if it is easily doable
* In the Visual column, I'd reverse the colors: green for negative and red for positive
  * Because for all quantities smaller is better
* For `nAlloc` and `nDealloc` I'd use the fixed-point formatting instead of scientific notation
  * I find it easier to see order-of-magnitude differences in that way
* It would be nice if changing "Quantity" would not reset the "Transition" to "total"
* "Abs(diff) high to low" does not seem to work correctly on other quantities than `added`
image

makortel:
Thanks for the update, looks really good (especially the presentation of the EventSetup records). As a last thing the EDModules now seem to show the "event setup" transition with zeroes
image
Could that be removed?

makortel:
Another last thing, sorry. Could all numbers in the tables be aligned to the right, and have at most (or exactly?) one digit after the period?

comparisons/module_alloc_monitor_viewer.html
```diff
@@ -0,0 +1,1417 @@
+<!doctype html>
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width,initial-scale=1">
+  <title>Module Allocation Monitor Viewer</title>
+  <link rel="preconnect" href="https://fonts.googleapis.com">
+  <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+  <link href="https://fonts.googleapis.com/css2?family=Space+Grotesk:wght@400;500;700&family=IBM+Plex+Mono:wght@400;500&display=swap" rel="stylesheet">
```
makortel: Are these links only for the font? Maybe we could just use a font the browser provides automatically?

makortel (commented):
The result looks good! I have one comment on the implementation.

comparisons/module_alloc_monitor_viewer.html
gartung: I guess so. But the font names are used throughout.

gartung: I can see if CoPilot can remove the use of this font.

gartung: Fonts and references to those fonts removed in the latest commit.

gartung:
test parameters:

workflows_profiling = 18634.21
enable = profiling

gartung:
please test

makortel:
Looks good to me

gartung:
@smuzaffar @akritkbehera please merge

smuzaffar:
+externals

PR #2732, opened 2026-04-29
moduleAllocMonitor diff table: remove unused Record column. Add additional metrics

### LINKED ISSUE #2192: Remove Record column from moduleAllocMonitor diff table. Right align IB, PR, diff column headers. Add additional metrics.
### Description

_No response_

gartung:


gartung:
test parameters:

workflows_profiling = 18634.0
enable = profiling

gartung:
please test with https://github.com/cms-sw/cmssw/pull/50822

pr_testing/run-pr-profiling.sh
```diff
@@ -122,7 +122,12 @@ EOF
       continue
     fi
     $CMS_BOT_DIR/comparisons/moduleAllocMonitor-circles-diff.py $CMSSW_VERSION-$BASENAME $f >$f.log || true
-    echo "<td><a target=\"_blank\" href=\"${PROFILING_WORKFLOW}/diff-$BASENAME.html\">diff-$BASENAME</a></td>" >> $WORKSPACE/upload/profiling/index-$PROFILING_WORKFLOW.html || true
+    echo "<td>" >> $WORKSPACE/upload/profiling/index-$PROFILING_WORKFLOW.html || true
+    if grep "Error: input files describe different metrics" $f.log ; then
+        echo "IB and PR files describe different metrics, comparing to self for diff<BR>" >> $WORKSPACE/upload/profiling/index-$PROFILING_WORKFLOW.html || true
```
makortel, line 127: It would be kind of nice to be able to deal with different sets of transitions in the diff, but maybe such situations will be rare enough that the effort for that is not really worth it.

makortel:
Where is the order of the transitions in the dropdown (and in the per-module tables) specified?
image
It would be nice if the order would be the same as in
https://github.com/cms-sw/cmssw/blob/982ab6323e0bf18367878f8eff42f07021be4945/PerfTools/AllocMonitor/scripts/edmModuleAllocJsonToCircles.py#L13-L33
It seems to me the orders in these two places are close to each other, but not quite the same.

gartung:
The extra transitions were discovered and added to the drop-down list. I will update to set the correct order.

pr_testing/run-pr-profiling.sh
gartung: I agree

gartung:
please test with https://github.com/cms-sw/cmssw/pull/50822

gartung:
please test with https://github.com/cms-sw/cmssw/pull/50822

pr_testing/run-pr-profiling.sh
gartung: Can this PR be merged and the different sets of transitions be solved in another PR?

makortel: > the different sets of transitions be solved in another PR?

I think we agreed the "different set of transitions" would not be worth of addressing.

makortel:
@cmsbuild, please test

pr_testing/run-pr-profiling.sh
gartung: If the code can handle missing modules it should be able to handle missing metrics without too much effort.

gartung: But for now I have reverted the branch to the original state.

makortel:
Looks good

mandrenguyen:
+1

mandrenguyen:
merge

PR #2733, opened 2026-05-01
Add timeout to command options in script

smuzaffar:


makortel:
This seems to conflict with `--maxmem_profile` option that we use in PR tests, see https://github.com/cms-sw/cmssw/pull/50292#issuecomment-4361166329

PR #2734, opened 2026-05-01
[Maybe merge?] Revert "Add timeout to command options in script"

makortel:
Reverts cms-sw/cms-bot#2733 at least to allow testing while it is figured out if cms-sw/cms-bot#2733 should really be reverted, or something else made instead.


smuzaffar:
+externals

PR #2748, opened 2026-05-13
chore: remove DQM signing permissions from Dimitris Papagiannis

gabrielmscampos:
Dimitris is no longer part of the team.

FYI @rseidita


smuzaffar:
+externals

PR #2754, opened 2026-05-18
[do not merge] Tesing use of external build resources stats

smuzaffar:


smuzaffar:
no needed any more as https://github.com/cms-sw/cms-bot/pull/2760 already have updated config.map for few non-production Ibs to use externals build stats

PR #2765, opened 2026-05-26
Update PR test workflows for 20_0_X

### LINKED ISSUE #2229: Update extra matrix workflows in cms-bot for 20_0_X
### Description

_No response_

makortel:
This PR updates the PR test workflow settings for 20_0_X that will break the backwards compatibility for non-RAW files. This means that workflows reading non-RAW files produced with <= 17_0_X will likely fail, so this PR disables those.

The profiling tests use workflows with pileup, so until new MinBias files become available, the best we can do is to use no-pileup workflow. Given the 20_0_X focus on phase 2, I think it is not worth of running Run 3 workflow for profiling by default there.

The GPU workflows have similar problem. We could use no-pileup workflows there too, but I want to confirm that explicitly first.

I _believe_ this PR can be merged already before the 20_0_X is opened.

Resolves https://github.com/cms-sw/framework-team/issues/2229

### DIFF
--- a/cmssw-pr-test-config
+++ b/cmssw-pr-test-config
@@ -24,7 +24,12 @@ if [ "${IS_EVOLUTION}" = "0" ]; then
   PR_TEST_MATRIX_EXTRAS=1306.0,101.0,9.0
 fi
 
-if [ "$CMSSW_VER" -ge 1601 ] ; then
+if [ "$CMSSW_VER" -ge 2000 ] ; then
+  # COMMENT: excluded workflows read non-RAW files produced with CMSSW < 20_0_X
+  PR_TEST_MATRIX_EXTRAS=1306.0,101.0,9.0
+  # COMMENT: use no-pileup workflow until first 20_X_Y MinBias files become available
+  PR_TEST_MATRIX_EXTRAS_GPU=34434.402,34434.403,34434.404,34434.712,34434.713,34634.7503,34634.751 #Phase2 Wfs (D121)
+elif [ "$CMSSW_VER" -ge 1601 ] ; then
   PR_TEST_MATRIX_EXTRAS_GPU=18634.402,18634.403,18634.406,18634.412,18634.422,18634.423 #2026 Wfs
   PR_TEST_MATRIX_EXTRAS_GPU=${PR_TEST_MATRIX_EXTRAS_GPU},34634.402,34634.403,34634.404,34634.712,34634.713,34634.7503,34634.751 #Phase2 Wfs (D121)
 elif [ "$CMSSW_VER" -ge 1600 ] ; then
@@ -49,7 +54,10 @@ if [ X"$PR_TEST_MATRIX_EXTRAS_GPU" != X"" ]; then
   done
 fi
 
-if [ "$CMSSW_VER" -ge 1601 ] ; then
+if [ "$CMSSW_VER" -ge 2000 ] ; then
+  # COMMENT: use no-pileup workflow until first 20_X_Y MinBias files become available
+  PR_TEST_MATRIX_EXTRAS_PROFILING=34434.21
+elif [ "$CMSSW_VER" -ge 1601 ] ; then
   PR_TEST_MATRIX_EXTRAS_PROFILING=34634.21,18634.21
 elif [ "$CMSSW_VER" -ge 1600 ] ; then
   PR_TEST_MATRIX_EXTRAS_PROFILING=34634.21,18634.21


cmssw-pr-test-config
```diff
@@ -49,7 +53,10 @@ if [ X"$PR_TEST_MATRIX_EXTRAS_GPU" != X"" ]; then
   done
 fi
 
-if [ "$CMSSW_VER" -ge 1601 ] ; then
+if [ "$CMSSW_VER" -ge 2000 ] ; then
+  # COMMENT: use no-pileup workflow until first 20_X_Y MinBias files become available
+  PR_TEST_MATRIX_EXTRAS_PROFILING=34434.21
```
makortel, line 59: FYI @gartung

cmssw-pr-test-config
```diff
@@ -24,7 +24,11 @@ if [ "${IS_EVOLUTION}" = "0" ]; then
   PR_TEST_MATRIX_EXTRAS=1306.0,101.0,9.0
 fi
 
-if [ "$CMSSW_VER" -ge 1601 ] ; then
+if [ "$CMSSW_VER" -ge 2000 ] ; then
+  # COMMENT: excluded workflows read non-RAW files produced with CMSSW < 20_0_X
+  PR_TEST_MATRIX_EXTRAS=1306.0,101.0,9.0
+  PR_TEST_MATRIX_EXTRAS_GPU=
```
makortel: @fwyzard We won't be able to run pileup workflows in 20_X_Y until new MinBias dataset(s) is/are produced. In the meantime, would it be useful to use the no-pileup workflows in the PR tests?

How long should 2026 workflows be kept here? The no-pileup workflows should continue to function until Run 3 reconstruction/simulation gets to be cleaned up (whose details are unclear ATM).

gartung (approved):

cmssw-pr-test-config
fwyzard: Thanks @makortel.

For the GPU workflows, I would rather run those without pileup than nothing.

As I expect that this will happen again in the next couple of years, at least until the Phase-2 raw data format stabilises, can we come up with a more flexible approach?

Thinking out loud - how many minbias events do we need to run the pileup workflows?
Can we teach the bot to generate them on demand, publish them on cvmfs, and update the workflows to pick the "current" sample automatically?

makortel: I added now the no-pileup phase 2 workflows.

> As I expect that this will happen again in the next couple of years, at least until the Phase-2 raw data format stabilises, can we come up with a more flexible approach?

In the leading order I'd think the MinBias GEN-SIM data types (for classical mixing) and the PREMIX data types (for premixing) would be independent of the Phase-2 RAW. Am I missing something?

> Thinking out loud - how many minbias events do we need to run the pileup workflows?

Depends on the amount of recycling one is willing to tolerate. I can only comment the obvious range between 1 and `N(PU) * N(BX) * N(signal events)` (IIRC 7 bunch crossings would be used Phase 2 simulation), but maybe @cms-sw/simulation-l2 would have better feeling how much of that could be brought down.

> Can we teach the bot to generate them on demand, publish them on cvmfs, and update the workflows to pick the "current" sample automatically?

I vaguely recall this option was discussed in the context of https://github.com/cms-sw/cmssw/issues/49795, but I leave it to @smuzaffar to comment more.

There is also a connection to  and https://github.com/cms-sw/cmssw/issues/49776

kpedro88: For MinBias recycling, it depends on the use case. If we just want technical tests of code and don't care about the physics output or about exploring more potential edge cases with a broader distribution of particles, one could just reuse 10 events over and over again. But as indicated, this might not be comprehensive.

https://github.com/cms-sw/cmssw/issues/49776 is still on my todo list.

fwyzard: > In the leading order I'd think the MinBias GEN-SIM data types (for classical mixing) and the PREMIX data types (for premixing) would be independent of the Phase-2 RAW. Am I missing something?

I was too approximate - I meant what affects the content of the raw data: geometry, electronic stimulation, etc.

makortel: > > In the leading order I'd think the MinBias GEN-SIM data types (for classical mixing) and the PREMIX data types (for premixing) would be independent of the Phase-2 RAW. Am I missing something?
> 
> I was too approximate - I meant what affects the content of the raw data: geometry, electronic stimulation, etc.

Thanks for the clarification, for those I could certainly imagine backwards-incompatibilities to come up (like with https://github.com/cms-sw/cmssw/issues/49795).

smuzaffar:
Looks good, this can go in. We have 20.0.x release cycle open. 20.0.x ib will be available soon

PR #2768, opened 2026-05-28
Increase MaxMemoryPreload error threshold to 30 MB for >= 17_0_X

makortel:
Following https://github.com/cms-sw/cmssw/issues/46966#issuecomment-4565013883

One-two workflows show random variations up to 25.4 MB. It is frequent enough to be annoying, so let's silence the warnings until the cause is understood.

Resolves https://github.com/cms-sw/framework-team/issues/2256


makortel:
FYI @gartung

makortel:
@cmsbuild, please test

smuzaffar:
+externals

PR #2773, opened 2026-05-31
Override docker image to May 19th

smuzaffar:
Add case handling for Docker image tags based on architecture.


smuzaffar:
please test workflow 180.1,181.1,182.1,183.1

PR #2775, opened 2026-06-01
Add PR comparison for Module Event Alloc Monitor circles plots

gartung:


gartung:
enable profiling

gartung:
please test with https://github.com/cms-sw/cmssw/pull/51095

comparisons/module_event_alloc_monitor_viewer.html
```diff
@@ -0,0 +1,1421 @@
...
+      "clear event",
+      "global end lumi",
+      "global end run",
+      "global write lumi",
+      "global write run",
+      "write process block",
+      "end stream",
+      "end job",
+      "destruction",
+    ];
```
makortel: These look like ModuleAllocMonitor-specific things that do not apply to `ModuleEventAllocMonitor`'s output.

gartung: Yes. This was left over from reusing the template page for ModuleAllocMonitor.

makortel:
Looks ok. This PR was tested via https://github.com/cms-sw/cmssw/pull/51095#issuecomment-4626543974

smuzaffar:
please test for CMSSW_17_0_X

@gartung @makortel , will this change work for old release cycles (i.e. releses without https://github.com/cms-sw/cmssw/pull/51095. change) ?

gartung:
The Module Event Alloc monitor script is not available in older releases. The profiling driver script will skip this test is the script is not available.

smuzaffar:
+externals

PR #2776, opened 2026-06-01
Add min as selectable aggregation to maxmem diff summary

gartung:


gartung:
enable profiling

makortel:
The behavior of "min" and "max" look ok. The "max of abs" is a bit weird because it changes the negatives to positives, which is exactly as requested, but is not very useful. Would be possible to show the original (positive or negative) values in the tables, and do the sorting based on "max of abs"? If that gets too weird, we could drop the "max of abs". The "min" is already an improvement.

gartung:
please test

makortel:
Looks reasonable

makortel:
@smuzaffar This PR can be merged

smuzaffar:
+externals

PR #2783, opened 2026-06-09
Make el9_amd64_gcc13 as production arch for 20.1.X release cycle

smuzaffar:
@cms-sw/orp-l2 this PR changes the production arch for CMSSW_20_1_X release cycle to be el9_amd64_gcc13. All the experimental IBs e..g DEVEL, ROCM, CLANG, GEANT4 etc. are also now use el9_amd64_gcc13 as their build arch


smuzaffar:
+externals

PR #2786, opened 2026-06-10
Update alca/db L2 with Elton

perrotta:
Replace Atanu with Elton as alca/db L2


smuzaffar:
+externals

requested by @cms-sw/alca-l2 @cms-sw/db-l2

PR #2787, opened 2026-06-10
PR Profiling: Add link for diff of IB and PR Vtune reports and add script to generate page with the diff of IB and PR sorted-CPU-RES reports

gartung:


gartung:
enable profiling

makortel:
I see links to the IB and PR VTune reports but not the diff
image

How about moving the VTune link between the FastTimerService and ModuleAllocMonitor? Then the time-based tools would be next to each other.

gartung:
Need to wait for the next IB to generate the vtune top-down json file before adding the comparison scripting.

makortel:
The links in "Event Loop Report Comparison" row do not work correctly.

makortel:
Given there is room, could these quantities
image

be fully spelled out? (especially the `pct` is not immediately clear)

The visualization is a bit confusing
image
image

The "visual" bar on the right goes from middle to the right for "increased", but from left to right for "decreased". How about "left to middle" for "decreased"?

How useful is the "Status" column? It should be rather obvious from the 3 Delta columns.

I guess the "Status" and "Visual" are from the "Delta % total"? Apparently it is possible to have negative "Delta total time" but positive "Delta % total". I wonder which one would be clearer to summarize as increase or decrease. As the main customer is @cms-sw/reconstruction-l2, I'd ask for their opinion.

gartung:
Status also indicates whether a workflow was only in IB or only in PR when the status drop down menu is set to all statuses. I defaulted it to in IB and PR. I will change the default for status and remove the visual column.

cuda/install-cuda.py
```diff
@@ -99,12 +99,10 @@ def fill(self, catalog, key, os_arch):
...
-        if 'cuda_variant' in component:
-          if len(component['cuda_variant'])>1:
-            raise RuntimeError(
-              f"the '{name}' component has multiple variants."
-            )
-          package = package['cuda%s' % component['cuda_variant'][0]]
+        if "cuda_variant" in component:
+            if len(component["cuda_variant"]) > 1:
+                raise RuntimeError(f"the '{name}' component has multiple variants.")
+            package = package["cuda%s" % component["cuda_variant"][0]]
```
makortel: Why is this file changed?

makortel:
I find it a bit weird to see "only in PR" / "only in IB" entries in the table for a PR that does not touch CMSSW. Does this perhaps imply these modules are so fast that VTune's sampling might miss them?

cuda/install-cuda.py
smuzaffar: @makortel , it was updated due to formatting reason. I had pushed this file yesterday to deploy cuda 13.3.0 under /cvmfs/project.cern.ch ( for CUDART IBs) but I think I did not run formatting. So I guess @gartung rebased his branch based on my changes and then bot ask to fix the formatting

gartung: I merged master into my branch and then got the formatting error.

gartung:
> I find it a bit weird to see "only in PR" / "only in IB" entries in the table for a PR that does not touch CMSSW. Does this perhaps imply these modules are so fast that VTune's sampling might miss them?

I will have to look into it.

gartung:
I need to update the IB and PR scripts to save a compressed copy of the source top-down txt report that the top-down json files are derived from. Then I can see if the script to produce the top-down json file is performing as expected.

cuda/install-cuda.py
gartung: I rebased to the commit before the change to cuda/install-cuda.py. @smuzaffar can you update the head of master with the formatting change?

smuzaffar: @gartung , formatting fixed for master

gartung:
@makortel I generalized the script that generates the doEvent summary to include all produce functions. I had copilot look at all produce functions and find the parent function. It found that "EventSetup Producers" were not included in the parent function list. This should lead to no missing produce functions in the diff report unless a new producer was added/removed in the PR.

makortel:
I think the EventSetup producers should not be added in "RECO event loop". The EventSetup producers are not run on every event, but on IOV boundaries (which is likely once per job in all test jobs).

gartung:
OK. Adding them did not make the number of IB vs PR functions reported the same. I will remove them from the report.

gartung:
please test

makortel:
Looks reasonable to me. @cms-sw/reconstruction-l2 Could you also take a look and give feedback if the diff report would suit your needs?

smuzaffar:
> Looks reasonable to me.

lets get this in then

PR #2789, opened 2026-06-13
Take first match if multiple IBs are matched for PR tests

smuzaffar:
If multiple IBs are matched for default PR tests then always take the first matching IB. Order in config.map decides which IB/arch will be used when one does `please test`


smuzaffar:
please test with cms-sw/cmsdist#10630

this should now use CMSSW_20_1_X/el9_amd64_gcc15 for tests instead of failing and reporting that no IB matched

smuzaffar:
+externals

as expected bot properly used `cms/2789/el9_amd64_gcc15 Waiting for status to be reported — Building CMSSW externals` for tests

smuzaffar:
abort

PR #2810, opened 2026-07-09
fix finding old externals size/file count

smuzaffar:
changes are needed due to update in toolfile format in 17.0.X and above IBs


smuzaffar:
please test with cms-sw/cmsdist#10711

smuzaffar:
+externals

PR #2814, opened 2026-07-17
Drop FP_X as with GCC13 we build every thing with frame pointers

smuzaffar:
As for GCC13 , we [enabled frame pointers](https://github.com/cms-sw/cmsdist/pull/9901) by default , so no need to have dedicated FP_X (build with frame-pointers enabled) IBs


smuzaffar:
+externals

PR #2832, opened 2026-09-03
Add bdanzi to Tracking POG and mmasciov to Trigger

mmasciov:
As per title:
- adding @bdanzi (Brunella D'Anzi) to Tracking POG category and group, as incoming Tracking POG L2;
- adding mmasciov (myself) to Trigger (as incoming HLT scouting L2).

### DIFF
--- a/categories.py
+++ b/categories.py
@@ -160,6 +160,7 @@ CMSSW_L2 = {
     "mbluj": ["tau-pog"],
     "mmasciov": ["tracking-pog"],
     "elusian": ["tracking-pog"],
+    "bdanzi": ["tracking-pog"],
     # PPD
     "abenecke": ["ppd"],
     "vlimant": ["ppd"],
--- a/groups.yaml
+++ b/groups.yaml
@@ -5,6 +5,7 @@ tracking-pog:
   - VourMa
   - mmasciov
   - elusian
+  - bdanzi
 btv-pog:
   - andrzejnovak
   - AlexDeMoor
--- a/watchers.yaml
+++ b/watchers.yaml
@@ -1757,3 +1757,7 @@ salimcerci:
 - SimDataFormats
 - Configuration/Skimming
 - DPGAnalysis/Skims
+mmasciov:
+- RecoTracker
+- Validation/RecoTrack
+- HLTrigger/Configuration/python/HLT_75e33


categories.py
```diff
@@ -40,6 +40,7 @@
             "silviodonato",
             "slava77",
             "srimanob",
+            "mmasciov",
```
smuzaffar: @mmasciov , this is not needed. You are l2 of @cms-sw/tracking-pog-l2 so you should be able to trigger PR tests. This `TRIGGER_PR_TESTS` contains the list of special users who are not in L2 list but we want them to trigger PR tests

categories.py
```diff
@@ -160,6 +161,7 @@
     "mbluj": ["tau-pog"],
     "mmasciov": ["tracking-pog"],
     "elusian": ["tracking-pog"],
+    "bdanzi": ["tracking-pog"],
```
smuzaffar, line 163: So @bdanzi is new @cms-sw/tracking-pog-l2 ?

categories.py
bdanzi: @mmasciov is the outgoing L2 for tracking POG, so he probably needs to be included in this special list.

categories.py
```diff
@@ -160,6 +161,7 @@
     "mbluj": ["tau-pog"],
     "mmasciov": ["tracking-pog"],
```
bdanzi, line 161: Maybe, to avoid any confusion, @mmasciov could remove yourself from this list in another commit? Otherwise, I can do it in a follow-up PR.

categories.py
mmasciov: I should indeed be replaced by Brunella in that L2 team.

categories.py
mmasciov: To my understanding, this list is not the same as the @cms-sw/tracking-pog-l2 team.

smuzaffar:
@mmasciov , the `TRIGGER_PR_TESTS` is not a list of `CMS HLT/Trigger` coordinators. So please do not add your self there.

smuzaffar:
+externals

mmasciov:
> @mmasciov , the `TRIGGER_PR_TESTS` is not a list of `CMS HLT/Trigger` coordinators. So please do not add your self there.

Done. However, my change was to be able to trigger PR tests (being a L2 in HLT/Trigger involved in development, it would be good for me to keep that ability, provided that eventually I'll lose those rights once I'll be removed from the tracking POG L2 group).

PR #2834, opened 2026-09-04
Add the new truth-graph packages

felicepantaleo:
Two packages of the MC-truth graph work are not yet in the map: https://github.com/cms-sw/cmssw/pull/51829

`SimGeneral/TruthGraphAssociatorProducers` holds the associators that match reconstructed objects to the truth graph. `Validation/TruthInfo` holds their DQM validation. Both come from the same development as `PhysicsTools/TruthInfo` and `SimDataFormats/TruthInfo`, which are already under `generators`, so they join the same category.

`Validation/TruthInfo` is also listed under `dqm`.

### DIFF
--- a/categories_map.py
+++ b/categories_map.py
@@ -720,6 +720,7 @@ CMSSW_CATEGORIES = {
         "Validation/TrackerHits",
         "Validation/TrackerRecHits",
         "Validation/TrackingMCTruth",
+        "Validation/TruthInfo",
     ],
     "externals": [
         "",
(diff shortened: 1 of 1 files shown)


felicepantaleo:
@cmsbuild please test

smuzaffar:
@felicepantaleo , can you please reference the cmssw PR where these packages are added?

smuzaffar:
ah I see https://github.com/cms-sw/cmssw/pull/51829

smuzaffar:
@cms-sw/dqm-l2 @cms-sw/generators-l2 any objection on adding these packages under dqm/generators?

gabrielmscampos:
+dqm
- No issues from our side.

sensrcn:
+generators

PR #2836, opened 2026-09-08
use first only major,minor and patch version fo config tag

smuzaffar:


smuzaffar:
please test for CMSSW_20_1_DEVEL_X

PR #2837, opened 2026-09-10
Replace 'ftorrresd' with 'enibigir' for xpog

tvami:
Emery is the incoming XPOG replacing Felipe


tvami:
@smuzaffar there is no need for running tests for this, right?

smuzaffar:
> @smuzaffar there is no need for running tests for this, right?

right, there is no need to run bot tests for such changes

PR #2839, opened 2026-09-11
Add VertexGNNReco to categories map

prabhatsolanki:
New package `DataFormats/VertexGNNReco`, introduced in cms-sw/cmssw#51862 (vertex-slot GNN primary-vertex reconstruction for Phase-2). It holds the Alpaka portable SoA collections of the track features and of the GNN output, so it is added to the `reconstruction` and `heterogeneous` categories, next to `DataFormats/VertexReco` and `DataFormats/VertexSoA`.


smuzaffar:
@cms-sw/heterogeneous-l2 @cms-sw/reconstruction-l2 any objections on adding this package `VertexGNNReco` under your responsibility ?

Moanwar:
+1

makortel:
The `DataFormats/VertexGNNReco` in https://github.com/cms-sw/cmssw/pull/51862 looks rather straightforward so I don't see a particular reason for it to be assigned to `heterogeneous` (not all alpaka-using `DataFormats` packages are assigned to `heterogeneous`). If others in @cms-sw/heterogeneous-l2 really want it, I don't object.

fwyzard:
Mostly I wonder if we need a new package at all...

fabiocos:
in principle ```DataFormats/VertexSoA``` is a place where the new data format could go without inflating the number of packages, a different SoA wrt the already existing one, but always related to the same reconstruction task

fabiocos:
actually, looking at the dependencies, ```DataFormats/VertexReco``` is perhaps preferable and more natural as a location

PR #2840, opened 2026-09-11
Add SoA data formats for FTLDigi and FTLRecHit

fabiocos:
In support of CMSSW PR cms-sw/cmssw#51861, SoA data formats for FTLDigi and FTLRecHit


smuzaffar:
@fabiocos , @cms-sw/upgrade-l2 is a legacy category so I don't think we should add any package under it. I suggest to add these under simulation.

@cms-sw/simulation-l2 @cms-sw/reconstruction-l2 any objections adding these package(s) under your responsibility ?

fabiocos:
@smuzaffar I agree that if upgrade is no more a supported category it is useless to continue adding things there. ```FTLDigiSoA``` is already under simulation, and ```FTLREcHitSoA``` under reconstruction, so that should be enough

smuzaffar:
+externals

PR #2844, opened 2026-09-16
Impement version switching inside process-pull-request, version heuristics and version checks in process_pr_v* 

iarspider:
@smuzaffar as discussed. Please check if any other environment variables needs to be preserved when retriggering the cms-bot job (https://github.com/cms-sw/cms-bot/compare/sign-hashes-not-commits...version_switching?expand=1#diff-7c7bddbc6ef2ecf51f761e670c6b0fb996a274d8a20f2c1ff6c22b1706c450c9R1192, https://github.com/cms-sw/cms-bot/compare/sign-hashes-not-commits...version_switching?expand=1#diff-e39247ebfb8d4717a2973aa953043137384efb7fad248b8d4d660ad2555412c5R6520)

I have updated process_pr.py from master branch to avoid possible merge conflicts when adding retriggering code.


process_pr.py
```diff
@@ -1155,6 +1179,29 @@ def get_combined_status_list(gh, last_commit, repository):
...
 
+def recreate_cms_bot_test_properties(bot_version: int = 2) -> None:
+    """
+    Create properties file to re-run cms-bot job with correct bot version
+
+    Args:
+        bot_version: Version number
+    """
+
+    params = {"CMS_BOT_VERSION": bot_version}
```
smuzaffar, line 1192: @iarspider , I think we only need `CMS_BOT_VERSION` and for the rest we configure jenkins job to use the existing parameters.

process_pr_v1.py
```diff
@@ -11,6 +11,7 @@
     EXTERNAL_REPOS,
 )
 from categories import CMSSW_CATEGORIES as default_CMSSW_CATEGORIES
+from process_pr_v2 import CMS_BOT_VERSION
```
smuzaffar: why we need this?

iarspider: Thanks for spotting

process_pr.py
```diff
@@ -102,8 +102,9 @@ def is_closed_branch(*args):
 def format(s, **kwds):
     return s % kwds
 
-
-BOT_CACHE_TEMPLATE = {"emoji": {}, "signatures": {}, "commits": {}}
+CMS_BOT_VERSION = 1
+BOT_CACHE_TEMPLATE = {"emoji": {}, "signatures": {}, "commits": {}, "version": CMS_BOT_VERSION}
+CMSSW_BRANCHES_FOR_AUTO_CODE_CHECKS = ["master", "CMSSW_17_0_X"]
```
smuzaffar, line 108: @iarspider , I suggest to not rename it. There are other tests and scripts which imports it so instead of renaming it fix the `process-pull-request` to use `process_pr` if version is `1`

iarspider: Sure.

smuzaffar:
@iarspider , what should we to with the process_pr tests? Looks like we need to keep all the `process_pr*.py` so that correct bot could be use to process the old PRs. In future if we add `v3` then we have to keep `v2` tests and new `v3` test working ... right? Should we just create `tests/v2`  directory for version 2 tests and in future new `vN` for newer `N`  version ?

iarspider:
> @iarspider , what should we to with the process_pr tests? Looks like we need to keep all the `process_pr*.py` so that correct bot could be use to process the old PRs. In future if we add `v3` then we have to keep `v2` tests and new `v3` test working ... right? Should we just create `tests/v2` directory for version 2 tests and in future new `vN` for newer `N` version ?

That's an option. We should also have a plan to remove tests for legacy versions once we stop updating them.

smuzaffar:
Overall I am fine with this PR to e merged in to new bot branch but feel free to update/move tests for v2 under dedicated directory

smuzaffar:
Or if you have better option for managing tests while keeping old bots working then please do propose

iarspider:
> Overall I am fine with this PR to e merged in to new bot branch but feel free to update/move tests for v2 under dedicated directory

I would prefer to do it in sign-hashes branch, or in a different temporary branch on top of it.

Let's wait for jenkins tests to finish, then I'll merge.

process-pull-request.py
```diff
@@ -81,13 +83,21 @@ def main():
...
 
     import repo_config
-    from process_pr_v2 import process_pr
 
     if not getattr(repo_config, "RUN_DEFAULT_CMS_BOT", True):
         return
     if getattr(repo_config, "REQUEST_PROCESSOR", "cms-bot") != "cms-bot":
         return
 
+    version = os.getenv("CMS_BOT_VERSION", 1)
```
smuzaffar: @iarspider , in case `CMS_BOT_VERSION` is set then it will return a string ... right? so you might want to use `int(....)` here

PR #2846, opened 2026-09-17
Update categories.py

antoniovagnerini:
Replacing outgoing L2 OV convener Adriano di Florio with Matheus Macedo


antoniovagnerini:
please test

smuzaffar:
no need to test such change. There is nothing bot can test here :-)

I am fine with merging it as long as @teteumac is correct github user and is new L2

